PR #6, opened 2020-05-23
Added note about Fedora kernel updates and compatibility

mskarbek:
To address comments like in this issue: https://github.com/openzfs/zfs/issues/10257#issuecomment-626018525

### DIFF
--- a/docs/Getting Started/Fedora.rst
+++ b/docs/Getting Started/Fedora.rst
@@ -12,14 +12,15 @@ released. Packages are available for the following configurations:
 | **Fedora Releases:** 30, 31, 32
 | **Architectures:** x86_64
 
-Take note that due to the release cycle of OpenZFS and Fedora adoption
-of new kernels in current distribution release, it may happen that you
-won't be able to build DKMS package on the most recent kernel update.
-For example, Fedora 32 was released with kernel 5.6 but it is expected
-that it will update to 5.7. OpenZFS was released with support for kernel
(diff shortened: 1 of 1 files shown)


docs/Getting Started/Fedora.rst
```diff
@@ -12,6 +12,15 @@ released. Packages are available for the following configurations:
...
 | **Architectures:** x86_64
 
+Take note that due to the release cycle of OpenZFS and Fedora adoption
+of new kernels in current distribution release, it may happen that you
+won't be able to build DKMS package on the most recent kernel update.
+For example, Fedora 32 was released with kernel 5.6 but it is expected
+that it will update to 5.7. OpenZFS was released with support for kernel
+5.6 but not for 5.7 and there will be no update right after 5.7 will hit
+Fedora repository so you have to take into account that you will have to
+pin your kernel version at some point.
```
gmelikov, line 22: Let's wrap it in a Note block:
```suggestion
.. note::
   Due to the release cycle of OpenZFS and Fedora adoption
   of new kernels in current distribution release, it may happen that you
   won't be able to build DKMS package on the most recent kernel update.
   For example, Fedora 32 was released with kernel 5.6 but it is expected
   that it will update to 5.7. OpenZFS was released with support for kernel
   5.6 but not for 5.7 and there will be no update right after 5.7 will hit
   Fedora repository so you have to take into account that you will have to
   pin your kernel version at some point.
```

PR #9, opened 2020-05-26
Add missing word

ahrens:
I think this meant to say "Welcome to the OpenZFS index" (per the above heading, "OpenZFS Index"), but maybe a different noun would be more appropriate here?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,7 +1,7 @@
-OpenZFS index
-=============
+OpenZFS Documentation
+=====================
 
-Welcome to the OpenZFS. This resource provides documentation for
+Welcome to the OpenZFS Documentation. This resource provides documentation for
 users and developers working with (or contributing to) the OpenZFS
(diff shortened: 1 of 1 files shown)


rlaager:
This should probably be "OpenZFS Documentation" both here and in the header?

docs/index.rst
```diff
@@ -1,7 +1,7 @@
 OpenZFS index
```
gmelikov, line 1:
```suggestion
OpenZFS Documentation
```

docs/index.rst
```diff
@@ -1,7 +1,7 @@
...
 =============
 
-Welcome to the OpenZFS. This resource provides documentation for
+Welcome to the OpenZFS index. This resource provides documentation for
```
gmelikov, line 4:
```suggestion
Welcome to the OpenZFS Documentation. This resource provides documentation for

PR #14, opened 2020-05-29
Set absolute paths in 404 page

gmelikov:
Signed-off-by: George Melikov <mail@gmelikov.ru>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,7 +12,7 @@ jobs:
     - uses: actions/checkout@v1
     - uses: ammaraskar/sphinx-action@master
       with:
-        pre-build-command: "apt-get update -y && apt-get install -y python3-pip && pip install -r docs/requirements.txt"
+        pre-build-command: "apt-get update -y && apt-get install -y git python3-pip && pip install -r docs/requirements.txt"
         docs-folder: "docs/"
     - name: Deploy
       uses: peaceiris/actions-gh-pages@v3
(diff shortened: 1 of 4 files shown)


rlaager:
If you're going to use something described as a hack anyway, the module clearly isn't the right choice for the job. Given that this is a trivial problem to solve and adding that module as a dependency is an issue for Debian (see #15), this is not the right answer.

I pushed a different fix (so I could get the GitHub Action to build it), which may or may not be the right answer either. I ultimately reverted it because while it worked locally, it didn't work when the GitHub Action built it:
```
diff --git a/docs/Makefile b/docs/Makefile
index 41dcdbb..289bcaf 100644
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -13,8 +13,13 @@ help:
 
 .PHONY: help Makefile
 
+# The 404 page can be served at any directory level, so its links must be
+# relative to the root.  Also, this command must be idempotent, as 404.html is
+# not rebuilt if it already exists and 404.rst was not changed.
 html: Makefile
        @$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+       @sed -Ei 's/(action|href|src)="([^\/#])/\1="\/\2/;s/\/http/http/' \
+               "$(BUILDDIR)/html/404.html"
 
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
```

gmelikov:
@rlaager unfortunately, we need to use this extension, it changes not only one HTML page, but ToC too. I've added a fix in it to work well with github pages and now it's not a "hack".

PR #15, opened 2020-05-29
README: all deps can be installed only via pip

gmelikov:
Signed-off-by: George Melikov <mail@gmelikov.ru>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,20 +12,14 @@ Building Locally
 Install Prerequisites
 ~~~~~~~~~~~~~~~~~~~~~
 
-Debian 11 (“testing”) / Ubuntu 20.04 or later::
+::
 
-   sudo apt install python3-sphinx python3-sphinx-issues python3-sphinx-rtd-theme
-
-Debian 10 (”Buster”) and earlier::
-
+   # For Debian based distros
    sudo apt install pip3
-   pip3 install -r requirements.txt
-   # Add ~/.local/bin to your $PATH, e.g. by adding this to ~/.bashrc:
-   PATH=$HOME/.local/bin:$PATH
+   # For RPM-based distros
+   sudo yum install python3-pip
 
-Other distros::
-
-   pip install -r requirements.txt
+   pip3 install -r docs/requirements.txt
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -12,20 +12,14 @@ Building Locally
...
+::
 
+   # For Debian based distros
    sudo apt install pip3
-   pip3 install -r requirements.txt
-   # Add ~/.local/bin to your $PATH, e.g. by adding this to ~/.bashrc:
-   PATH=$HOME/.local/bin:$PATH
-
-Other distros::
+   # For RPM-based distros
```
mskarbek, line 19: There are ready to use rpm packages in Fedora - `python3-sphinx`, `python3-sphinx-issues`, `python3-sphinx_rtd_theme`.

gmelikov: There are some modules that don't exist in distro packages, that's why I think it's easier to just install requirements via `pip install -r requirements.txt`

gmelikov: Specifically, it's module from https://github.com/openzfs/openzfs-docs/pull/14

mskarbek: There is also a package for that module in Fedora `python3-sphinx-notfound-page.noarch`. :)

gmelikov: @mskarbek ah, Fedora :) I looked at base centos repo.

Main point - it depends :) So better to give at least one 99% working solution.

gmelikov: I may bring some other modules soon, let's get back to this queston in a month to have a minimally settled dependencies as a start, otherwise it may be a pointless time. 

Maybe it would be easier to work in venv, so we may just add one `make me happy` command to prepare venv, deps and generate docs there.

rlaager: @mskarbek
> There are ready to use rpm packages in Fedora - `python3-sphinx`, `python3-sphinx-issues`, `python3-sphinx_rtd_theme`.

I'm not finding these on: https://src.fedoraproject.org From that site, it looks like maybe they don't have the "3":
```
  sudo yum install python-sphinx python-sphinx-issues \
      python-sphinx_rtd_theme
```

rlaager:
Merged

Issue #11, opened 2020-05-29
Ubuntu 20.04 Troubleshooting section should include commands for mounting encrypted rpool


randomphrase:
I am a bit of a beginner with ZFS, so apologies for any incorrect terminology.

If you have followed the process for installing an encrypted rpool at installation time by modifying the installer as described in the overview section, you subsequently need to use `zfs load-key rpool` (or something similar) before mounting it. The troubleshooting section should include the correct command, whatever it is.

ghfields:
Just want to be clear... this isn't a problem with using a modified installer.  Anyone trying to use the troubleshooting section who has native encryption on their rpool will face this issue.

Submitted https://github.com/openzfs/openzfs-docs/pull/16

rlaager:
I fixed this in d1b3973b92af1ea2105d42f1786231a82a38e179 but apparently my syntax for closing the issue was wrong. (I think it needed the # on the issue number.)

PR #16, opened 2020-05-29
Fix missing native decrypt in Troubleshooting

ghfields:
Fixes https://github.com/openzfs/openzfs-docs/issues/11
Signed-off-by: Garrett Fields <ghfields@gmail.com>


rlaager:
I use `zpool load-key -a` in the Debian Buster version, so I copied that over. `zfs load-key` is a no-op if there are no encrypted filesystems. I'm open to changing the `-a` to `rpool` if people think that is useful. They're the same if you follow the HOWTO and do nothing else. If you create other datasets that you don't have the keys for, then `-a` is a problem. If you create other datasets that you need mounted in the rescue environment and you do have the keys, then `rpool` is a problem.

Issue #2, opened 2020-05-21
Add man pages


gmelikov:
- [X] Add man pages
- [X] Integrate CSS
- [ ] Parse links to existing pages and add hyperlinks on them
- [ ] Differ pages per version/platform
- [ ] Change "Edit on github" for generated pages into something more appropriate than "page not found"
- [ ] Trigger rebuild on man pages changes in `/openzfs/zfs`
- [ ] Review and fix syntax in man pages

PR #20, opened 2020-06-02
Add man pages

gmelikov:
- original pages are converting into HTML by `mandoc` itself
- script integrates resulting HTML into rst structure
- special CSS file added for a pretty view
- sphinx static search works for man pages too in this case
- interesting side effect: HTML preview of man pages can show flaws in original syntax, so any non-CSS problems with man pages should be fixed in original file

Preview: https://gmelikov.github.io/openzfs-docs/man/


.github/workflows/publish.yml
```diff
@@ -10,9 +10,17 @@ jobs:
...
+        sudo apt-get update -y
+        sudo apt-get install -y git python3-pip mandoc
+        pip install -r docs/requirements.txt
+    - name: Gen_man_pages
+      run: make man
+    - name: Gen_sphinx
+      uses: ammaraskar/sphinx-action@master
       with:
-        pre-build-command: "apt-get update -y && apt-get install -y git python3-pip && pip install -r docs/requirements.txt"
+        pre-build-command: "apt-get update -y && apt-get install -y git"
```
rlaager, line 23: It looks like git is installed earlier already. Likewise below in pull_request.yaml.

docs/Makefile
```diff
@@ -20,3 +20,16 @@ html: Makefile
...
+ZFS_GIT_DIR = ./_build/zfs
+
+.PHONY: man
+man:
+	if [ ! -d $(ZFS_GIT_DIR) ]; \
+		then git clone --depth=1 $(ZFS_GIT_REPO) ./_build/zfs; \
+	else \
+		git -C $(ZFS_GIT_DIR) pull ; \
+	fi
+	../scripts/man_pages.py ./_build/zfs/man ./
```
rlaager, line 35: Missing end-of-file newline.

docs/_static/css/mandoc.css
```diff
@@ -0,0 +1,251 @@
...
+	text-align: right;
+}
+.foot td {
+	padding: 1em;
+}
+
+/* Sections and paragraphs. */
+
+.manual-text {
+		margin-left: 0em; }
```
rlaager, line 39: There is some goofy formatting (of the CSS itself) / whitespace starting here.

scripts/man_pages.py
```diff
@@ -0,0 +1,143 @@
...
+
+man_sections = {
+    '1': 'User Commands',
+    '2': 'System Calls',
+    '3': 'C Library Functions',
+    '4': 'Devices and Special Files',
+    '5': 'File Formats and Conventions',
+    '6': 'Games et. al.',
+    '7': 'Miscellanea',
+    '8': 'System Administration tools and Daemons',
```
rlaager, line 33: capitalize Tools. Or, "System Administration Commands"

scripts/man_pages.py
```diff
@@ -0,0 +1,143 @@
...
+import subprocess
+
+
+man_sections = {
+    '1': 'User Commands',
+    '2': 'System Calls',
+    '3': 'C Library Functions',
+    '4': 'Devices and Special Files',
+    '5': 'File Formats and Conventions',
+    '6': 'Games et. al.',
```
rlaager, line 31: I'd drop the "et. al" here.

scripts/man_pages.py
```diff
@@ -0,0 +1,143 @@
...
+    '7': 'Miscellanea',
```
rlaager, line 32: I'd probably call this "Miscellaneous" as `man man` does.

rlaager:
I see you linked to the output. Yeah, at first glance, the output looks great too!

.github/workflows/publish.yml
gmelikov: Unfortunately, `ammaraskar/sphinx-action@master` doesn't use that, looks like it runs in different container with same artifacts. Or maybe I did something wrong.

docs/_static/css/mandoc.css
gmelikov: It's an upstream formatting, i'd prefer to leave it as is to easily backporting in future, added same comment.

docs/Makefile
```diff
@@ -20,3 +20,16 @@ html: Makefile
...
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+# Gen man pages
+ZFS_GIT_REPO = https://github.com/openzfs/zfs.git
+ZFS_GIT_DIR = ./_build/zfs
+
+.PHONY: man
+man:
+	if [ ! -d $(ZFS_GIT_DIR) ]; \
+		then git clone --depth=1 $(ZFS_GIT_REPO) ./_build/zfs; \
```
behlendorf, line 31: nit: can you move this `then` to the previous line, after the `;`, as is traditional.

docs/Makefile
```diff
@@ -20,3 +20,16 @@ html: Makefile
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+# Gen man pages
+ZFS_GIT_REPO = https://github.com/openzfs/zfs.git
+ZFS_GIT_DIR = ./_build/zfs
```
behlendorf, line 26: Always generating the man page from master is the right first step.  But I think we should consider early on how we're going to handle generating documentation for multiple versions of OpenZFS.  Adding a "Docs by version" section to the sidebar as was done for the [Python documentation](https://docs.python.org/3/) would be a nice way to handle this.  Let's make sure to leave ourselves some options since I'm not sure how difficult it would be to add this kind of thing latter.

ahrens:
image

It looks like there's a rendering issue where this appears to say `zfssetproperty=value` when it should be `zfs set property=value`.  Seems that all similar places have the same issue.

ahrens:
This looks great!  Could we consider highlighting the "main" manpages, e.g. zfsconcepts, zfs, zpool?  They kind of get lost among the subcommands and things that are not as relevant to sysadmins (e.g. fsck.zfs, zdb).

ahrens:
I noticed that when another manpage is mentioned, it is a HTML link, but clicking on it doesn't seem to do anything.  Is that just an issue with the preview, or do we need to fix something so that it will work?

gmelikov:
Thanks everybody for the feedback, i'll address it soon!

@ahrens making links active is in my plan, you can see it in issue https://github.com/openzfs/openzfs-docs/issues/2 and comment what we might see there too.

docs/Makefile
gmelikov: You're right, I thought about it too. Looks like it would be pretty easy, because now we generate docs. We may use one of sphinx modules for that, for ex. https://github.com/sphinx-contrib/sphinxcontrib-versioning to get versioned docs + man pages via script addtions.

Worst thing that may get us is different links and the need to manage redirects. In this case I prefer some 404 links from other sites but we get faster publication -> better docs now.

gmelikov:
>It looks like there's a rendering issue where this appears to say zfssetproperty=value when it should be zfs set property=value. Seems that all similar places have the same issue.

It was chrome-specific, fixed, thank you for a cathing that!

PR #21, opened 2020-06-03
Add Fedora Root On ZFS guide and in general revamp Fedora's wiki

cheesycod:
@rlaager Add a Fedora Root On ZFS guide to the wiki. Note that this guide is still in draft and should not be used until all bugs are finished. Happy to have it reviewed though.
Signed-off-by: Rootspring <cheesycod@outlook.com>


cheesycod:
Note, this is a draft
I accidentally made it a full PR

cheesycod:
Can someone give me feedback?

cheesycod:
@ahrens can you review this

gmelikov:
@cheesycod thank you for a contribution!
Unfortunately, we don't have a maintainer who uses Fedora with root on ZFS, so we can't review it fast. We need a one, are you ready to support it in future?

Maybe we can generalize howto somehow to use same parts in debian/ubuntu/fedora? It make the process easier.

Feel free to use other howtos as a base and trigger me when you think the instruction is stable enough and tested.

docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
```diff
@@ -51,7 +51,8 @@ Contributing
 
 #. Install the tools::
 
-    sudo apt install pip3
+    sudo apt install pip3 sensible-utils
```
gmelikov, line 54: Why do we need `sensible-utils`? I prefer to have a minimal requirements to start.

cheesycod: Because we call sensible-browser, which is a part of sensible-utils. I think this is a part of Debian by default though

rlaager: It should be installed, but even if it's not, it's not a big deal. The point is to show where the index.html is. I just use `sensible-browser` because it's a real command that will do the right thing. But it's not really necessary to install it.

gmelikov: What do you think to move it into new `optional` section then?

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,818 @@
...
+deduplication is a permanent change that cannot be easily reverted.
+
+Support
+~~~~~~~
+
+If you need help, reach out to the community using the :doc:`zfs-discuss
+mailing list <../../Project and Community/Mailing Lists>` or IRC at
+`#zfsonlinux <irc://irc.freenode.net/#zfsonlinux>`__ on `freenode
+<https://freenode.net/>`__. If you have a bug report or feature request
+related to this HOWTO, please `file a new issue and mention @rlaager
```
rlaager, line 45: I don't use Fedora. Take my name out. If you're going to maintain this, you can sub in your username. Also, update the URL below for filing an issue, which has my username in it and also has the distro name and version (so if someone clicks the link and keeps the suggested text, I know which HOWTO we are talking about).

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,818 @@
...
+#. Make your changes.
+
+#. Test::
+
+    cd docs # If you aren't already in the docs folder
+    make html
+    sensible-browser _build/html/index.html
+
+#. ``git commit --signoff`` to a branch, ``git push``, and create a pull
+   request. Mention @rlaager.
```
rlaager, line 70: Likewise I'm mentioned here.

cheesycod:
I'm ready to support it in the future. Fedora Root On ZFS is what I use on all my PC's.
These instructions are not yet stable and needs to be tested.
> Maybe we can generalize howto somehow to use same parts in debian/ubuntu/fedora? It make the process easier.
Yes, this would be good, right now, I copied the Debian Buster Root On ZFS guide and modified it for Fedora.
About grub, I still haven't gotten Fedora's build of grub to work yet and I don't really know if I want to support grub given that it doesn't really work unless I rebuild it from source code and use ZPOOL_VDEV_NAME_PATH=1. I'll most likely just add some special side notes at the end for people who really want to use grub.

rlaager:
I've been line-wrapping these at the usual 79 characters (to leave one for a diff). You might consider doing the same. It makes the diffs easier to view in GitHub.

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
cheesycod: Yeah, I was just focusing on writing the guide. I'll fix this now

rlaager:
> I'm ready to support it in the future. Fedora Root On ZFS is what I use on all my PC's.
> These instructions are not yet stable and needs to be tested.
> 
> > Maybe we can generalize howto somehow to use same parts in debian/ubuntu/fedora? It make the process easier.
> 
> Yes, this would be good, right now, I copied the Debian Buster Root On ZFS guide and modified it for Fedora.

I'm not in love with this. I realize it's a pain to update the HOWTOs when there is copied-and-pasted text all over, but there is a good reason that there are separate files. Going from distro to distro or even version to version, there will be changes required. Some are small and some are big. Some are critical, and some are not. But if you want something that is step-by-step, it really has to be tailored to each distro version. To keep the maintenance burden down, my practice is to only maintain the current version for each distro. I don't backport changes to the previous versions. I used to delete those HOWTOs, but people asked that I leave them up for reference.

> About grub, I still haven't gotten Fedora's build of grub to work yet and I don't really know if I want to support grub given that it doesn't really work unless I rebuild it from source code and use ZPOOL_VDEV_NAME_PATH=1. I'll most likely just add some special side notes at the end for people who really want to use grub.

In general, I recommend you follow the distro defaults as closely as possible, unless there is a compelling reason otherwise. If Fedora doesn't use GRUB, then don't both with GRUB.

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
rlaager: No worries. I understand it's a draft.

cheesycod:
> In general, I recommend you follow the distro defaults as closely as possible, unless there is a compelling reason otherwise. If Fedora doesn't use GRUB, then don't bother with GRUB.

For Fedora, GRUB is a lot harder to support. Fedora's grub doesn't work out of the box with ZFS while systemd-boot is pretty well integrated into Fedora and just works without hassle through kernel updates. That's why I haven't added full GRUB support yet.

cheesycod:
OK, I am going to run a full bare metal runthrough using this guide so I can catch mistakes

cheesycod:
I meant to say fix number 1

docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
cheesycod: OK, I removed sensible-utils

cheesycod:
> > I'm ready to support it in the future. Fedora Root On ZFS is what I use on all my PC's.
> > These instructions are not yet stable and needs to be tested.
> > > Maybe we can generalize howto somehow to use same parts in debian/ubuntu/fedora? It make the process easier.
> > 
> > 
> > Yes, this would be good, right now, I copied the Debian Buster Root On ZFS guide and modified it for Fedora.
> 
> I'm not in love with this. I realize it's a pain to update the HOWTOs when there is copied-and-pasted text all over, but there is a good reason that there are separate files. Going from distro to distro or even version to version, there will be changes required. Some are small and some are big. Some are critical, and some are not. But if you want something that is step-by-step, it really has to be tailored to each distro version. To keep the maintenance burden down, my practice is to only maintain the current version for each distro. I don't backport changes to the previous versions. I used to delete those HOWTOs, but people asked that I leave them up for reference.

I agree with this. Fedora needs a lot of steps that Debian and Ubuntu simply do not need and a lot of things for different distros need different install procedures

ahrens:
@cheesycod I'm not really familiar enough with Fedora to add a useful review, sorry.

docs/Getting Started/Fedora/index.rst
```diff
@@ -0,0 +1,64 @@
...
+.. contents:: Table of Contents
+  :local:
+
+Installation
+------------
+
+If you want to use ZFS as your root filesystem, see the `Root on ZFS`_
+links below instead.
+
+Currently, due to Fedora's unstable fast-paced kernel ABI (kABI), only DKMS style packages can be provided at this time.
```
rlaager, line 15: "Currently" and "at this time" mean the same thing. Pick one or the other. :)

docs/Getting Started/Fedora/index.rst
```diff
@@ -0,0 +1,64 @@
...
+As of right now, the following configurations are supported:
+  
+**Fedora Releases**: 30, 31, 32
```
rlaager, line 18: Fedora 30 does not seem to be supported upstream (i.e. by Fedora) any more. If that is the case, please remove it here and below.

docs/Getting Started/Fedora/index.rst
```diff
@@ -0,0 +1,64 @@
...
+Note that it is important to make sure that the matching *kernel-devel*
+package is installed for the running kernel since DKMS requires it to
+build ZFS.
+
+.. code:: sh
+
+   $ sudo dnf install kernel-devel zfs
+
+If the Fedora provided *zfs-fuse* package is already installed on the
+system. Then the ``dnf swap`` command should be used to replace the
```
rlaager, line 53: These should be one sentence: "If the ... system, the ``dnf swap`` command ..."

docs/Getting Started/Fedora/index.rst
```diff
@@ -0,0 +1,64 @@
...
+existing fuse packages with the ZFS on Linux packages.
```
rlaager, line 54: s/fuse/FUSE/

docs/Getting Started/Fedora/index.rst
```diff
@@ -0,0 +1,64 @@
...
+| **Location**: /etc/pki/rpm-gpg/RPM-GPG-KEY-zfsonlinux
+| **Fedora 30 Package**: http://download.zfsonlinux.org/fedora/zfs-release.fc30.noarch.rpm
+| **Fedora 31 Package**: http://download.zfsonlinux.org/fedora/zfs-release.fc31.noarch.rpm
+| **Fedora 32 Package**: http://download.zfsonlinux.org/fedora/zfs-release.fc32.noarch.rpm
+| **Download from**: pgp.mit.edu
+| **Fingerprint**: C93A FFFD 9F3F 7B03 C310 CEB6 A9D5 A1C0 F14A B620
+
+.. code:: sh
+
+   $ sudo dnf install http://download.zfsonlinux.org/fedora/zfs-release$(rpm -E %dist).noarch.rpm
```
rlaager, line 37: I don't have a Fedora system handy. Does `rpm -E %dist` output `.fc32` or just `fc32`. If it's the latter, you need a `.` in front of the `$(...)`.

docs/Getting Started/Fedora/index.rst
```diff
@@ -0,0 +1,64 @@
...
+   To simplify installation a zfs-release package is provided which includes a zfs.repo configuration file and the ZFS on Linux public signing key. All official ZFS on Linux packages are signed using this key, and by default both yum and dnf will verify a package’s signature before allowing it to be installed. Users are strongly encouraged to verify the authenticity of the ZFS on Linux public key using the fingerprint listed here.
+.. warning::
+   Firstly, note that Fedora Rawhide is not officially supported and things will break if you use it.
+   Also that the OpenZFS team is not responsible for any damages caused by using ZFS On Linux and that the latest code on GitHub is not tested and may break your system.
+
+| **Location**: /etc/pki/rpm-gpg/RPM-GPG-KEY-zfsonlinux
```
rlaager, line 28: This bit seems to duplicate the code below. Do we really need both anyway? I'd just keep the code.

docs/Getting Started/Fedora/index.rst
mskarbek: `.fc32`

docs/Getting Started/Fedora/index.rst
cheesycod: Yeah, I put that here because Fedora 30 was still in the wiki for Fedora

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
cheesycod: It should be fixed now.

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
cheesycod: Ok, fixed

docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
cheesycod: ...Is this resolved. Should I close it?

docs/Getting Started/Fedora/index.rst
cheesycod: Wait, should I remove it?

cheesycod:
Any more changes?

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,753 @@
...
+~~~~~~~
+
+- This HOWTO uses a whole physical disk.
+- Do not use these instructions for dual-booting.
+- Backup your data. Any existing data will be lost.
+
+System Requirements
+~~~~~~~~~~~~~~~~~~~
+
+- `64-bit Fedora Workstation (spins are not yet tested yet)
```
mskarbek, line 22: second "yet" unnecessary

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,753 @@
...
+  <https://getfedora.org/en/workstation/download/>`__
+- `A 64-bit kernel is strongly encouraged.
+  <https://openzfs.github.io/openzfs-docs/Project%20and%20Community/FAQ.html#bit-vs-64-bit-systems>`__
+- Installing on a drive which presents 4 KiB logical sectors (a “4Kn” drive)
+  only works with UEFI booting. This not unique to ZFS. `GRUB does not and
+  will not work on 4Kn with legacy (BIOS) booting.
+  <http://savannah.gnu.org/bugs/?46700>`__
+- You MUST use a UEFI system for this guide as GRUB is not supported and will not be supported in the forseeable future. To verify that you are using UEFI, go to /sys/firmware/efi/efivars and make sure it is not empty.
```
mskarbek, line 30: It's worth to note that Fedora developers are discussing dropping legacy BIOS support entirely in future releases: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/QBANCA2UAJ5ZSMDVVARLIYAJE66TYTCD/

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,753 @@
...
+       sgdisk     -n3:0:0        -t3:BF00 $DISK
+
+   - LUKS (same warning as with Unencrypted and ZFS native encryption, change the -n3 and -t3 to -n2 and -t3 if you are not adding swap)::
+
+       sgdisk     -n3:0:0        -t3:8309 $DISK
+
+   If you are creating a mirror or raidz topology, repeat the partitioning
+   commands for all the disks which will be part of the pool.
+
+#. Create the root pool (change the disk to ${DISK}-part2 if you did):
```
mskarbek, line 206: if you did what?

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,753 @@
...
+           -O encryption=aes-256-gcm \
+           -O keylocation=prompt -O keyformat=passphrase \
+           -O acltype=posixacl -O canmount=off -O compression=lz4 \
+           -O dnodesize=auto -O normalization=formD -O relatime=on \
+           -O xattr=sa -O mountpoint=/ -R /mnt \
+           rpool ${DISK}-part2 # ${DISK}-part2 if swap partition was not created
+
+   - LUKS::
+
+       apt install --yes cryptsetup
```
mskarbek, line 232: `apt`?

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,753 @@
...
+
+     zfs create -o com.sun:auto-snapshot=false  rpool/var/lib/nfs
+
+   A tmpfs is recommended later, but if you want a separate dataset for
+   ``/tmp``::
+
+     zfs create -o com.sun:auto-snapshot=false  rpool/tmp
+     chmod 1777 /mnt/tmp
+   
+   Note that the reason why we are not fully seperating everything like we did in Ubuntu is because dnf will fail to install or update certain packages if we create too many datasets.
```
mskarbek, line 378: can you add a comment with an example of such package

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,753 @@
...
+   filesystem. It also allows you to set a quota on ``rpool/tmp``, if you want
+   to limit the maximum space used. Otherwise, you can use a tmpfs (RAM
+   filesystem) later.
+
+#. Copy the LiveCD to your HDD/SDD::
+
+     rsync -avxHASX / /mnt/
+   
+   It is important to not forget the trailing /.
+   This command copies the LiveCD to our new zfs datasets and this is the only way I have found to reliably install and boot Fedora Workstation
```
mskarbek, line 396: what issues did you have during the installation of the `Fedora Workstation` packages group?

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,753 @@
...
+
+#. Configure the network interface:
+
+   Find the interface name::
+
+     ip addr show
+
+   Adjust ``NAME`` below to match your interface name::
+
+     vi /mnt/etc/network/interfaces.d/NAME
```
mskarbek, line 425: it is 2020, it's time to stop using old config files and either use `nmconnection` in `/etc/NetworkManager/system-connections/` when using NetworkManager which is default or `network` unit files in `/etc/systemd/network` when using `systemd-networkd`

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,753 @@
...
+QEMU/KVM/XEN
+~~~~~~~~~~~~
+
+Set a unique serial number on each virtual disk using libvirt or qemu
+(e.g. ``-drive if=none,id=disk1,file=disk1.qcow2,serial=1234567890``).
+
+To be able to use UEFI in guests (instead of only BIOS booting), run
+this on the host::
+
+  sudo apt install ovmf # or dnf install edk2-ovmf (if the host is Fedora)
```
mskarbek, line 721: `apt`?

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,753 @@
...
+initrd is unlikely to contain sensitive data.) The system cannot boot
+without the passphrase being entered at the console. Performance is
+good. As the encryption happens in ZFS, even if multiple disks (mirror
+or raidz topologies) are used, the data only has to be encrypted once.
+
+LUKS encrypts almost everything. The only unencrypted data is the bootloader,
+kernel, and initrd. The system cannot boot without the passphrase being
+entered at the console. Performance is good, but LUKS sits underneath ZFS, so
+if multiple disks (mirror or raidz topologies) are used, the data has to be
+encrypted once per disk. Note that I (cheesycod) do not use LUKS and that you are using this at your own risk.
```
gmelikov, line 95: This is an official project documentation, let's not use ` I ` statements. And, if we are not sure on anything - we may not want to write about it.

Fox ex, you can write here:
```suggestion
encrypted once per disk. 

.. note::
This manual describes LUKS configuration only as an example, it wasn't tested on everyday basis usage.
```

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
cheesycod: Yes, basesystem/filesystem (I forgot which one, need to check) will fail if those datasets are present.

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
cheesycod: System fails to boot and hangs at Starting GNOME Desktop Manager

cheesycod: TTY also fails at this point and doesn't allow you to login

cheesycod:
Are there any other changes needed?

gmelikov:
@cheesycod is it already tested?

I'll try look at it after august 10.

cheesycod:
Yes, I just tested it

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,751 @@
...
+     <https://github.com/zfsonlinux/zfs/commit/82a37189aac955c81a59a5ecc3400475adb56355>`__.
+     Inside ZFS, extended attributes are used to implement POSIX ACLs.
+     Extended attributes can also be used by user-space applications.
+     `They are used by some desktop GUI applications.
+     <https://en.wikipedia.org/wiki/Extended_file_attributes#Linux>`__
+     `They can be used by Samba to store Windows ACLs and DOS attributes;
+     they are required for a Samba Active Directory domain controller.
+     <https://wiki.samba.org/index.php/Setting_up_a_Share_Using_Windows_ACLs>`__
+     Note that ``xattr=sa`` is `Linux-specific
+     <http://open-zfs.org/wiki/Platform_code_differences>`__. If you move your
```
gmelikov, line 294:
```suggestion
     <https://openzfs.org/wiki/Platform_code_differences>`__. If you move your
```

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,751 @@
...
+
+     zfs create -o com.sun:auto-snapshot=false  rpool/var/lib/nfs
+
+   A tmpfs is recommended later, but if you want a separate dataset for
+   ``/tmp``::
+
+     zfs create -o com.sun:auto-snapshot=false  rpool/tmp
+     chmod 1777 /mnt/tmp
+   
+   Note that the reason why we are not fully seperating everything like we did in Ubuntu is because dnf will fail to install or update certain packages if we create too many datasets. An example of one such package is filesystem, which fails to install if other ZFS datasets are created.
```
gmelikov, line 386:
```suggestion
   Note that the reason why we are not fully separating everything like we did in Ubuntu is because dnf will fail to install or update certain packages if we create too many datasets. An example of one such package is filesystem, which fails to install if other ZFS datasets are created.
```

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,751 @@
...
+   filesystem. It also allows you to set a quota on ``rpool/tmp``, if you want
+   to limit the maximum space used. Otherwise, you can use a tmpfs (RAM
+   filesystem) later.
+
+#. Copy the LiveCD to your HDD/SDD::
+
+     rsync -avxHASX / /mnt/
+   
+   It is important to not forget the trailing /.
+   This command copies the LiveCD to our new zfs datasets and this is the only way I have found to reliably install and boot Fedora Workstation
```
gmelikov, line 404:
```suggestion
   This command copies the LiveCD to our new zfs datasets and this is the only way guide's author found to reliably install and boot Fedora Workstation
```

gregory-lee-bartholomew:
## Under Step 2: Disk Formatting, 3. Partition your disk(s), I'd suggest:

1. Use +1G for the size of the ESP. There is a possibility that Root on ZFS users may want to keep more snapshots/clones and they may need to keep extra kernels and initrds on the ESP to support them.
2. You can use "0" instead of fixed partition numbers with sgdisk. It will automatically choose the next-available partition number. Doing so might help avoid a little confusion with the numbering if people skip creating the swap partition. Use 0 for all the options that are supposed to reference the same partition in one command (both -n and -t).
3. I like setting a label when creating the partitions (e.g. `-c 0:boot`) so that the output from commands like `sgdisk -p /dev/sda` will be a little more user friendly. This partlabel also shows up as a symlink under `/dev/disk/by-partlabel` and can be used in places like /etc/fstab.
4. If you go with suggestion 3, maybe note that the labels will need to be unique for each partition in the comment about creating a mirrored topology. E.g.:

>If you are creating a mirror or raidz topology, repeat the partitioning commands for all the disks which will be part of the pool. Also append a unique id to each partlabel (-c 0:boot.a, -c 0:boot.b, etc.).

## Under Step 3: System Installation, 3. Create datasets:

1. I would omit the following and leave /root on the root filesystem. I used to create a separate filesystem for /root, but I found that it can lead to a problem where, if you do any rescue operations under the dracut emergency shell (e.g. to regenerate the initrd), files may end up left under the /root mountpoint (e.g. /root/.bash_history) and then the /root filesystem will fail to mount from then on because the /root mountpoint is not empty.

```
zfs create -o mountpoint=/root             rpool/home/root
```

## Under Step 4: System Configuration, 3. Install systemd-boot:

I think you need to set a context for selinux to work properly. Setting the mode bits might also be a good idea. E.g.:

```
PARTLABEL=boot /boot vfat umask=0077,shortname=lower,context=system_u:object_r:boot_t:s0,nofail 0 0
````

Also, for mirrored disk setups, I have some scripts hosted on github that you can refer people to if they want to keep their /boot partition synchronized across the disks.

https://github.com/gregory-lee-bartholomew/bootsync

## Under Step 6: Fixing systemd-boot config:

I've found that a lot of the problems with the /boot/loader/entries files are caused by grub2 scripts that get triggered on kernel updates. A simple solution to the problem is to just remove and denylist all grub2 rpms. If you add the following commands just prior to installing systemd-boot in the chroot environment, you can probably eliminate this section.

```
rpm --nodeps -ve $(rpm -qa | grep "^grub2-") os-prober
mkdir /boot/$(</etc/machine-id)
echo 'exclude=grub2-*,os-prober' >> /etc/dnf/dnf.conf
```

Also, you should set the kernel parameters under /etc/kernel/cmdline for systemd-boot. E.g.:

```
echo 'root=zfs:pool/0 bootfs.snapshot quiet rhgb' > /etc/kernel/cmdline
```

## Under Step 7: First Boot:

You might add some notes about how to verify that the zfs kernel module is available for the right kernel version and to verify that it is present in the initramfs. I find the following particularly useful:

```
dkms status
lsinitrd /boot/$(</etc/machine-id)/$(uname -r)/initrd
```

As for the note about "If you updated your kernel ...". You might want to change your guide to ensure that that doesn't happen during the initial setup. Especially since there is a possibility that zfs may not work with the latest kernel. One way to do that is to replace `dnf ...` with `dnf --disablerep=* --enablerepo=fedora ...`. If you want to run `dnf update` and update everything except the kernel, you can replace that with `dnf update --exclude=kernel*`.

## Other suggestions.

1. I've found that my _hostid_ can sometimes end up being different between my LiveCD session and the final installation. A simple fix for this problem is to define a hostid **before** you first create the zpool in the LiveCD session. You can do this with the following command:

```
dd if=/dev/random of=/etc/hostid bs=1 count=4
```

The hostid is normally based on the current IP address, but it is really just meant to be a random unique identifier. If you are curious, you can use the following command to decode your current hostid:

```
hostid | perl -e '$/=\2; $,="."; $\="\n"; print map { eval "0x$_" } (<>)[1,0,3,2];'
```

File: docs/Getting Started/Debian/Debian Buster Root on ZFS.rst [hunk 22]
```diff
@@ -60,7 +61,7 @@ Contributing
 
 #. Test::
 
-    cd docs
+    cd docs # If you aren't already in the docs folder
```
rlaager, line 64: I think people can figure that one out.

docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
```diff
@@ -52,6 +52,7 @@ Contributing
 #. Install the tools::
 
     sudo apt install pip3
+    cd openzfs-docs/docs
```
rlaager, line 55: I've merged this idea (but in a different way) in 371d3cf08ba235e68b36df3285975b4ca450b7bd.

docs/Getting Started/Fedora/index.rst
```diff
@@ -0,0 +1,73 @@
...
+   To simplify installation of OpenZFS, a zfs-release package is provided. This package includes a zfs.repo configuration file and the ZFS on Linux public signing key. All official ZFS on Linux packages are signed using this key, and by default, dnf will verify a packages signature before allowing it to be installed unless specified otherwise using ``--nogpgcheck`` or through the ``gpgcheck`` configuration option in dnf.conf or in the repo. Users are strongly encouraged to not disable this feature and to verify the authenticity of the ZFS on Linux public key using the fingerprint listed here.
+.. warning::
+   The OpenZFS team is not responsible for any damages caused by using ZFS On Linux. By using this software, you are responsible for any and all data loss that occurs. While we may provide user support and bug reporting channels, we are not obliged to help you whatsoever. Please be cautious and understand what you are doing before using openZFS.
```
rlaager, line 25: openZFS -> OpenZFS (capitalize the first letter)

docs/Getting Started/Fedora/index.rst
```diff
@@ -0,0 +1,73 @@
...
+.. note::
+   Please note that on Fedora Rawhide, you MUST use the nodebug kernels. Note that modifying of the OpenZFS source code to spoof GPL (license violation in itself) is not supported either.
```
rlaager, line 27: I'd remove the "(license violation in itself)" piece. I'm not sure it's anywhere near settled that changing "CDDL" to "GPL" on your own system would be a license violation. It probably isn't, as long as you're not distributing anything, even under the most restrictive license interpretations. But more importantly, it really doesn't matter if it's a violation or not; it's not supported either way.

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
rlaager: This isn't so much an issue of _old_ config files as this is a Debian-ism that hasn't been ported to Fedora's way yet.

gregory-lee-bartholomew:
There are several places in the guide where you specify to operate on `${DISK}-part3` and then explain in comments that the value will need to be changed to `${DISK}-part2` if they skipped creating the swap partition. Since you are now setting labels on the partitions with sgdisk, you could provide the below instead of `${DISK}-part{2,3}` to automate finding the right partition.

```
/dev/disk/by-partuuid/$(blkid -t PARTLABEL=root -s PARTUUID -o value)
```

Using the above might make the guide easier to follow and eliminate a potential point of confusion for the reader. Just point out that the _partlabel_ needs to match what was provided to the **-c** parameter of sgdisk.

Arguably, partition UUIDs might be even more stable than device paths since paths can change if cables are shuffled around while adding drives or if partition numbers are changed with something like `sgdisk --sort /dev/sda`.

rlaager:
> There are several places in the guide where you specify to operate on `${DISK}-part3` and then explain in comments that the value will need to be changed to `${DISK}-part2` if they skipped creating the swap partition.

Partition numbers are assigned by the admin and need not be contiguous. This can be simplified by _always_ using partition number 3 (for both creation and use) regardless of whether swap is used or not. If there is no swap, there will be no partition 2, and that is totally fine.

gregory-lee-bartholomew:
Excerpted from `man sgdisk` (emphasis added):

>**-s, --sort**
Sort partition entries. GPT partition numbers need not match the order of partitions on the disk. If you want them to match,  you can  use  this  option. _Note that some partitioning utilities sort partitions whenever they make changes._ Such changes will be reflected in your device filenames, so you may need to edit /etc/fstab if you use this option.

If it is true that some partitioning utilities will re-sort the partition numbers, then I don't think it is a good idea to skip partition numbers (or to rely on them at all).

rlaager:
I’m not making a general case argument about whether to use UUID vs partition numbers. Obviously, the former is more stable for any number of reasons and should be used in persistent configurations, like `/etc/fstab`. I am only speaking to what the instructions should have the user manually type and run. For that, my view is that partition numbers are easier to work with.

In that context, it doesn’t really matter what some hypothetical tool might do. All that matters is what the instructions as written do.

gregory-lee-bartholomew:
FYI, just in case you are having trouble with the _kernel-install_ command, it looks like your instructions are running `rm -rvf /boot` a second time and the second instance is removing the `/boot/<machine-id>` directory. I think the kernel-install command keys on the presence of that subdirectory. So if it is not working for you, then that might be why.

cheesycod:
Is there a benifit to using kernel-install instead of reinstalling the kernel which also fixes any corrupt kernel installs etc.

gregory-lee-bartholomew:
> Is there a benifit to using kernel-install instead of reinstalling the kernel which also fixes any corrupt kernel installs etc.

It's up to you of course, but I would say that kernel-install is the right tool for the job. It should be much faster and only do what is needed whereas reinstalling requires uninstalling the existing kernel and all its drivers, redownloading the rpms, and then reinstalling and recompiling the needed drivers. Recompiling can take a particularly long time. If you use kernel-install, and the zfs driver is already built for the kernel that you are installing (as indicated by `dkms status`), then kernel-install will just use the already compiled driver and it will be much faster.

There should not be any corrupt kernel installs. Particularly since you have deleted everything under /boot. All the kernels and initramfs' that are used during startup are stored under /boot. So since you've deleted them all, there are no corrupt ones (I'm assuming that by corrupt, you mean initramfs' that do not have the zfs driver included).

Just my two cents.

cheesycod:
There is just one problem with kernel-install. We do a "dnf update" before doing anything in the chroot (fedora sometimes breaks their gpg keys or changes functionality etc.) so using uname -p would result in the older kernel being installed if it was ever updated

gregory-lee-bartholomew:
> There is just one problem with kernel-install. We do a "dnf update" before doing anything in the chroot ...

I would highly recommend replacing `dnf update` with `dnf update --exclude=kernel*` for this guide. It is very likely (inevitable really) that the guide will not work for future users if `dnf update` is used because it frequently happens that Fedora's latest kernel is incompatible with the currently-available zfs release.

If you don't update the kernel as part of the install process, then `uname -r` will return the correct version.

Again, these are just my personal recommendations.

cheesycod:
One question, does kernel-install automatically create the systemd-bootd config files

gregory-lee-bartholomew:
> One question, does kernel-install automatically create the systemd-bootd config files

Yes. Just be sure:

1. The /boot/<machine-id> directory exists.
2. The grub2-* rpms have been removed.
3. The desired kernel parameters are set in /etc/kernel/cmdline

I can set up a VM to test the whole procedure out when I have a bit more time.

cheesycod:
OK, I am testing that out

cheesycod:
Yep, @gregory-lee-bartholomew It works and the instructions have been updated to use kernel-install.

gregory-lee-bartholomew:
A quicker alternative to rebuilding the initramfs just in case would be to run instead:

`lsinitrd /boot/$(</etc/machine-id)/$(uname -r)/initrd | grep zfs.ko`

The above is what I always do. If **zfs.ko** is present in the initramfs, then you should be good to go. If not, then you would need to rebuild the initramfs. The kernel-install command on the previous line should have just run dracut, so it really shouldn't be necessary to run it a second time. kernel-install calls /usr/lib/kernel/install.d/50-dracut.install, if you are curious to see exactly how it calls dracut.

cheesycod:
Another thing, msdosfs/systemd-boot seems to enforce the partlabel to EFI. Also the partlabels seem to be made after mounting on my system so...

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -629,9 +629,9 @@ Step 8: Last Minute Fixes
 
      sed -i 's/#timeout.*/timeout 10/' /boot/loader/loader.conf
 
-#. Ensure that zfs-mount.service is enabled::
+#. Ensure that zfs-target is enabled::
 
-     systemctl enable zfs-mount.service
+     systemctl enable zfs.target
```
gregory-lee-bartholomew, line 634: I've recently encountered a Fedora system that failed to mount the zfs pools during boot because the zfs kernel module was not available in time. I found that adding `rd.driver.pre=zfs` to the list of kernel parameters resolved the problem. I suspect that may be what is needed here instead of `systemctl enable zfs.target`. The zfs target should be enabled by default.

Just FYI.

cheesycod: the zfs.target isnt to solve thats, its to solve the edge case in which the zfs pools and partitions arent loaded for some reason

cheesycod:
sorry for the inactivity, i was a bit busy

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
gregory-lee-bartholomew: If there is even a chance that the system could fail to boot due to sub-filesystems failing to mount, I'd be tempted to minimize the layout down to just / and /home (if /home fails to mount, it will not prevent the system from booting and it is easy to mount after booting using the root account). It looks like your guide is pretty close to doing that already. Also, I _think_ that is pretty close to how Fedora 33 is going to layout their default btrfs file system: https://fedoramagazine.org/btrfs-coming-to-fedora-33/#comment-502491

I see that your current version has /var/log and /var/spool as separate file systems. If they were to fail to mount, the system should still boot, but it might be good to note that if they fail once, they will continue to fail on subsequent boots/attempts if any files are left under those mountpoints. So it might be helpful to inform the user that they will need to delete auto-generated files under those directories to recover from the error. Or better yet, IMHO, just keep them on the / file system by default and let the user know that if they don't want their logs rolled back, then they will need to make /var/log a separate file system.

I still haven't tried it out, but your guide is looking prettying good so far. Thanks for doing the research on this.

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
```diff
@@ -0,0 +1,746 @@
...
+   ::
+
+     cp /usr/share/systemd/tmp.mount /etc/systemd/system/
+     systemctl enable tmp.mount
+
+.. note::
+   GRUB installation is not supported by this guide and will not be supported in the forseeable future. The above steps should have installed systemd-boot, an alternative to GRUB which provides the majority of GRUBS features. If you still wish to use GRUB, it might be possible to chainload systemd-boot using GRUB and boot your Fedora installation that way. Instructions on how to do this will not be provided and this has not been tested to work. You are welcome to make a PR for this however
+
+Step 6: First Boot
+------------------
```
Zaryob, line 513: Before the boot is it need zfs-list.cache files for rpool and bpool.

docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
Zaryob: +1

Zaryob (commented):
It seem no problem

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
Zaryob: I tried it without zfs-list.cache files for pools it was crashed...

docs/Getting Started/Fedora/Fedora Root On ZFS.rst
cheesycod: Your welcome

darkbasic:
It would be awesome to have something similar for RHEL as well!

gmelikov:
Thank you everyone, we merged instructions for Fedora in actual #177 , feel free to make changes via new PRs!

PR #24, opened 2020-06-05
Source list in live context

Arooo:
Line 104 : "/mnt/etc/apt/sources.list" should be  "/etc/apt/sources.list", we are preparing the live context, not the chroot context, yet.


rlaager:
Apparently I fixed this locally but forgot to commit and push. Thanks!

Arooo:
Thanks a lot for all of us ;)

Le 05/06/2020 à 23:27, Richard Laager a écrit :

PR #27, opened 2020-06-11
Fix Ubuntu 20.04 download link.

Aevin1387:
This fixes the Ubuntu 20.04 Root on ZFS docs pointing to the Desktop 18.04.3 ISO. I left as http though the link is also available over https. cc @rlaager 

Signed-off-by: Cory Stephenson <aevin@me.com>


rlaager:
Manually merged. I amended to use https. Good catch on both counts! Thanks!

rlaager:
For reference, this was commit 72bbf8d68e41fca8c70cb360c8809cd8efc318f5.

rlaager:
Oh wow, and then I screwed up the commit and didn't succeed in merging in the https. Doh! Fixed in 933a4c6d377ff039c213ef267666d6fe846228bc.

PR #30, opened 2020-06-17
Switch open-zfs.org links to openzfs.org, and use HTTPS URLs

jwatt:
The open-zfs.org host, Dreamhost, returns a page for https://open-zfs.org
(note, httpS://) which causes HTTPS Everywhere with the 'Encrypt All
Eligible Sites' option enabled to redirect http://open-zfs.org to
https://open-zfs.org, and therefore fail to redirect to openzfs.org.
Ideally a redirect would be implemented at Dreamhost for
httpS://open-zfs.org too, but presumably the links in the docs should
really be updated anyway.

### DIFF
--- a/docs/Developer Resources/Git and GitHub for beginners.rst
+++ b/docs/Developer Resources/Git and GitHub for beginners.rst
@@ -58,7 +58,7 @@ SunOS <https://www.cis.upenn.edu/~lee/06cse480/data/cstyle.ms.pdf>`__,
 `ZFS on Linux Developer
 Resources <https://github.com/zfsonlinux/zfs/wiki/Developer-Resources>`__,
 `OpenZFS Developer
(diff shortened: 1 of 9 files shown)


gmelikov:
Thank you!

Let's fix all open-zfs.org links then, I've found at least one another in https://github.com/openzfs/openzfs-docs/blob/master/docs/Project%20and%20Community/Mailing%20Lists.rst , please grep it.

jwatt:
It looked like for the mailing lists the original open-zfs.org domain was still in use (see https://openzfs.org/wiki/Newcomers for example). Can you clarify if that's no longer the case?

gmelikov:
I mean that link open-zfs.org/wiki/Mailing_list may be updated too here.

ahrens:
Thanks for fixing this.  I'll also see what I can do about the https://open-zfs.org redirect.

PR #33, opened 2020-06-26
Specify buster-backports usage for Debian Buster root installation

spotlightishere:
`apt` silently ignores the usage of the `-t buster-backports` flag when on the step of preparing the installation environment. Without buster-backports as a source, everything ZFS is pulled the buster repository. As of writing, this is ZFS 0.7.12, which does not support many of the features as specified during bpool creation, causing later creation failure.

(This also was mentioned in the [openzfs/zfs repo wiki](https://github.com/openzfs/zfs/wiki/Debian-Buster-Root-on-ZFS/4dd322f1182a7ed28d36acda691779bdd2ef682d), alongside other Debian wiki resources. I guess it jumped the migration!)

Signed-off-by: Spotlight <spotlight@joscomputing.space>


rlaager:
Thanks!

PR #35, opened 2020-07-15
Encrypted swap patch not necessary for tmpfs

jknockaert:
Just a small note clarifying the necessity of the patch. I didn't test this.


rlaager:
I'm not sure how `/tmp` is implicated here?

jknockaert:
Ja right. I meant that the patch is not necessary if one chooses not to enable swap (instead of ``tmp``). Sorry for the confusion. I updated the branch I created for this PR, but somehow this is not (yet) reflected in the PR. Not sure where it went wrong.

PR #36, opened 2020-07-20
edonr checksum is not secure enough for dedup

gmelikov:
Mark it as not appropriate.

Fixes https://github.com/openzfs/zfs/issues/10594


rlaager:
Shouldn’t this be something like: yes for nopwrite; dedup requires “verify”

Or split nopwrite and dedup into two columns?

gmelikov:
> Shouldn’t this be something like: yes for nopwrite; dedup requires “verify”
> 
> Or split nopwrite and dedup into two columns?

Added this info into notes, you're right!

PR #49, opened 2020-08-26
Migrate ZFS-* message descriptions

### LINKED ISSUE #3: Add ZFS messages
https://github.com/zfsonlinux/zfsonlinux.github.com/tree/master/msg

freqlabs:
Prerequisite for changing the URLs in the command outputs from
zfsonlinux.org to openzfs.github.io/openzfs-docs/

Closes #3

### DIFF
--- a/docs/_TableOfContents.rst
+++ b/docs/_TableOfContents.rst
@@ -8,4 +8,5 @@
    Performance and Tuning/index
    Basic Concepts/index
    man/index
+   msg/index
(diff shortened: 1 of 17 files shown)


behlendorf:
I think it would be nice as a follow up PR to change the illumos style disk names to what a FreeBSD or Linux user might expect to see.

gmelikov (approved):

PR #56, opened 2020-09-13
Move `Performance tuning` and `Hardware` pages from openzfs wiki

gmelikov:
Pandoc raw mediawiki->rst conversions are in separate commits for ease of review.

I'll add redirection in wiki after this PR merged.

We may want to move this pages because there are contributors ready to update them (there are some inconsistencies in Perf tuning page on wiki, for ex. with Postgresql case).

### DIFF
--- a/docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
@@ -37,8 +37,7 @@ deduplication is a permanent change that cannot be easily reverted.
 Support
 ~~~~~~~
 
-If you need help, reach out to the community using the :doc:`zfs-discuss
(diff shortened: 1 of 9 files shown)


gmelikov:
@ryao as the main contributor of these pages you may want to review this PR, thank you!

@ahrens hope you're not against moving these pages here, I've got some folks who may contribute to them.

rlaager (commented):
This looks good to me, with some notes:
1. nit: s/Open ZFS/OpenZFS/
2. The mention that only FreeBSD has TRIM support is out of date.
3. The note about Linux setting the noop schedule should be removed. That is either no longer done or is deprecated.

rlaager (approved):
LGTM

ahrens:
I'm fine with moving these to github.  Thanks!

PR #58, opened 2020-09-26
Update Fedora.rst

kenno:
Clarify that *zfs-testing* repository is only available for Fedora 31 and older. It is not available for Fedora 32.

### DIFF
--- a/docs/Getting Started/Fedora.rst
+++ b/docs/Getting Started/Fedora.rst
@@ -70,7 +70,7 @@ Testing Repositories
 --------------------
 
 In addition to the primary *zfs* repository a *zfs-testing* repository
-is available. This repository, which is disabled by default, contains
+is available for Fedora 31 or older. This repository, which is disabled by default, contains
 the latest version of ZFS on Linux which is under active development.
 These packages are made available in order to get feedback from users
 regarding the functionality and stability of upcoming releases. These


gmelikov:
This is a temporary issue https://github.com/openzfs/zfs/issues/10964 , maybe better to highlight that fedora updates are so fast that usually ZFS repo for new release will appear later?

kenno:
The minor modification is only to clarify the "Testing Repositories" section on that page, which wasn't available since Fedora 31. (I should have checked it before making the change in Fedora.rst.)

* http://download.zfsonlinux.org/fedora-testing/31/x86_64/repodata/repomd.xml

I think we should mention somewhere on that page that the **zfs-testing** repository is no longer available for Fedora 31 and newer unless there is plan to resurrect the Testing Repository for Fedora 32.

If we make it a bit clearer that **zfs-testing** repo is not available for Fedora 31, 32 (until further notice), then no Fedora users should come and report that this is repo is not working. (We can easily point the reporter to that doc page.)

I know it doesn't matter much, but sometimes it's easier to troubleshoot problems when we can eliminate other possibility.

behlendorf:
> unless there is plan to resurrect the Testing Repository for Fedora 32.

Indeed there is.  We'll see about getting it updated this week.

kenno:
I hope you do not take this request as a pressure to resurrect the **zfs-testing** repository. 

Thanks @gmelikov and @behlendorf for taking time to response to this concern, and thanks to all the devs of this project.

behlendorf:
The zfs-testing repository has been added, it contains the OpenZFS 2.0.0-rc3.

kenno:
> The zfs-testing repository has been added, it contains the OpenZFS 2.0.0-rc3.

Thank you so much. I'm using this 2.0 rc3 version with kernel 5.8.12-200.fc32. Working flawlessly.

PR #61, opened 2020-10-07
Correct references to the `zfsonlinux` org in developer docs

strugee:
I split the paragraph reformatting into a separate commit so the diff was easier to read but you'll probably want to squash it before merging. (Or I can do it.)

Also, the verbiage about "copy of the upstream OpenZFS code" still seems wrong but I wasn't sure exactly what was right. Is `openzfs/zfs` the repo of record now? Or is that still illumos?

### DIFF
--- a/docs/Developer Resources/Building ZFS.rst
+++ b/docs/Developer Resources/Building ZFS.rst
@@ -5,24 +5,23 @@ GitHub Repositories
 ~~~~~~~~~~~~~~~~~~~
 
 The official source for ZFS on Linux is maintained at GitHub by the
-`openzfs <https://github.com/openzfs/>`__ organization. The project
-consists of two primary git repositories named `spl
-<https://github.com/openzfs/spl>`__ and `zfs
-<https://github.com/openzfs/zfs>`__, both are required to build ZFS on
-Linux.
+`openzfs <https://github.com/openzfs/>`__ organization. The primary
+git repository for the project is the `zfs
+<https://github.com/openzfs/zfs>`__ repository.
 
-**NOTE:** The SPL was merged in to the
-`zfs <https://github.com/openzfs/zfs>`__ repository, the last major
-release with a separate SPL is ``0.7``.
-
--  **SPL**: The SPL is thin shim layer which is responsible for
-   implementing the fundamental interfaces required by OpenZFS. It's
-   this layer which allows OpenZFS to be used across multiple platforms.
+There are two main components in this repository:
 
 - **ZFS**: The ZFS repository contains a copy of the upstream OpenZFS
    code which has been adapted and extended for Linux and FreeBSD. The
    vast majority of the core OpenZFS code is self-contained and can be
    used without modification.
+
+- **SPL**: The SPL is thin shim layer which is responsible for
+   implementing the fundamental interfaces required by OpenZFS. It's
+   this layer which allows OpenZFS to be used across multiple
+   platforms. SPL used to be maintained in a separate repository, but
+   was merged into the `zfs <https://github.com/openzfs/zfs>`__
+   repository in the ``0.8`` major release.
 
 Installing Dependencies
 ~~~~~~~~~~~~~~~~~~~~~~~
@@ -67,11 +66,11 @@ depends on your requirements.
    downside to using packages it is greatly increases the time required
    to build, install, and test a change.
 
--  **In-tree**: Development can be done entirely in the SPL and ZFS
-   source trees. This speeds up development by allowing developers to
-   rapidly iterate on a patch. When working in-tree developers can
-   leverage incremental builds, load/unload kernel modules, execute
-   utilities, and verify all their changes with the ZFS Test Suite.
+- **In-tree**: Development can be done entirely in the SPL/ZFS source
+   tree. This speeds up development by allowing developers to rapidly
+   iterate on a patch. When working in-tree developers can leverage
+   incremental builds, load/unload kernel modules, execute utilities,
+   and verify all their changes with the ZFS Test Suite.
 
 The remainder of this page focuses on the **in-tree** option which is
 the recommended method of development for the majority of changes. See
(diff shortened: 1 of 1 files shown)


docs/Developer Resources/Building ZFS.rst
```diff
@@ -5,24 +5,24 @@ GitHub Repositories
...
-`zfsonlinux <https://github.com/zfsonlinux/>`__ organization. The
-project consists of two primary git repositories named
-`spl <https://github.com/zfsonlinux/spl>`__ and
-`zfs <https://github.com/zfsonlinux/zfs>`__, both are required to build
-ZFS on Linux.
+`openzfs <https://github.com/openzfs/>`__ organization. The project
+consists of two primary git repositories named `spl
+<https://github.com/openzfs/spl>`__ and `zfs
+<https://github.com/openzfs/zfs>`__, both are required to build ZFS on
+Linux.
```
ahrens, line 12: The `spl` repo hasn't been used for quite a while now, I think you can remove references to it (per the note below), and just address building 0.8 and later code.

strugee: @ahrens pushed a new commit removing all references to building <= 0.7. Sorry for the ugly diff :/

ahrens (approved):

docs/Developer Resources/Building ZFS.rst
```diff
@@ -5,24 +5,23 @@ GitHub Repositories
 ~~~~~~~~~~~~~~~~~~~
 
 The official source for ZFS on Linux is maintained at GitHub by the
```
gmelikov, line 7:
```suggestion
The official source for OpenZFS is maintained at GitHub by the
```

gmelikov (approved):
LGTM, later it would be great to separate build instructions for Linux and FreeBSD here.

docs/Developer Resources/Building ZFS.rst
strugee: Nice catch :+1: fixed another instance too in a followup commit

strugee:
Squashed the paragraph reformatting into the first commit.

docs/Developer Resources/Building ZFS.rst
```diff
@@ -4,25 +4,24 @@ Building ZFS
...
+There are two main components in this repository:
 
--  **ZFS**: The ZFS repository contains a copy of the upstream OpenZFS
-   code which has been adapted and extended for Linux. The vast majority
-   of the core OpenZFS code is self-contained and can be used without
-   modification.
+- **ZFS**: The ZFS repository contains a copy of the upstream OpenZFS
+   code which has been adapted and extended for Linux and FreeBSD. The
+   vast majority of the core OpenZFS code is self-contained and can be
+   used without modification.
```
gmelikov, line 17:
```suggestion
- **ZFS**: The OpenZFS repository contains a generic code and system-dependent
   parts for Linux and FreeBSD. The vast majority of the core OpenZFS code
   is self-contained and can be used without modification.
```

strugee: Thanks, wasn't sure what exactly was correct here. Typo: "contains generic code" instead of "contains a generic code"

docs/Developer Resources/Building ZFS.rst
```diff
@@ -4,25 +4,24 @@ Building ZFS
...
+git repository for the project is the `zfs
+<https://github.com/openzfs/zfs>`__ repository.
+
+There are two main components in this repository:
 
--  **ZFS**: The ZFS repository contains a copy of the upstream OpenZFS
-   code which has been adapted and extended for Linux. The vast majority
-   of the core OpenZFS code is self-contained and can be used without
-   modification.
+- **ZFS**: The ZFS repository contains a copy of the upstream OpenZFS
```
rlaager, line 14: In terms of eliminating the "copy of upstream", maybe something like this:

**ZFS**: The ZFS repository contains the OpenZFS code which supports FreeBSD and Linux. The vast majority of the OpenZFS code is OS-independent.

docs/Developer Resources/Building ZFS.rst
gmelikov: I like @rlaager 's wording more.

docs/Developer Resources/Building ZFS.rst
Ornias1993: @rlaager I agree "copy of upstream" should be gone. However:

"The vast majority of the OpenZFS code is OS-independent."

 What do you want to say here?
- The original meaning was the fact it was "self contained" and usable without modification.
- Your version doesn't seem to have the same meaning, it's rather vague.
- also "The ZFS repository contains the OpenZFS code which supports FreeBSD and Linux", is rather vague too... It could just as well be read "contains compatibility code" or "Contains code that supports FreeBSD and Linux" (without being clear it's the main repo for the project)

How about something like, to keep the original meaning in tact:
```suggestion
- **ZFS**: The ZFS repository contains the main OpenZFS code with support for FreeBSD and Linux. The vast majority of the OpenZFS code is OS-independent and can be used without modification.
```

rlaager: In terms of "self contained" vs "OS-independent", I actually have the same question: I have no idea what the original intention was.

If the audience of this is a user of ZFS, rather than someone porting it to a new platform, I'm not sure that we need the second sentence anyway.

Ornias1993: @rlaager You are right, I was getting confused by that second scentence., reading both versions this tells all we need:

```suggestion
- **ZFS**: The ZFS repository contains the OpenZFS code which supports FreeBSD and Linux.
```
_(or my version of the same scentence, thats more of a style choice)_

That being said:
Shouldn't there be a comma after code?
`The ZFS repository contains the OpenZFS code, which supports FreeBSD and Linux.`
vs
`The ZFS repository contains the OpenZFS code which supports FreeBSD and Linux.`

afaik one of those is:
"the code to supports something" and the other "the code having/including support for something"

behlendorf (approved):

gmelikov:
I've merged it as is, we can work out nits later.

PR #63, opened 2020-10-07
Add compatibility_matrix.py + Feature flags page

gmelikov:
Generated preview https://gmelikov.github.io/openzfs-docs/Basic%20Concepts/Feature%20Flags.html

compatibility_matrix.py origin is https://zgrep.org/zfs.py

Easier to preview by commit, first one is https://zgrep.org/zfs.py as is.

### DIFF
--- a/scripts/compatibility_matrix.py
+++ b/scripts/compatibility_matrix.py
@@ -14,18 +14,13 @@
 # 2018-07-05: Created.
 # ????-??-??: Many things happened.
 # 2020-02-23: Applied patch by rlaager.
-# 2020-02-30: Show domain prefixes (via Vlad Bokov), partially apply patch by rlaager.
+# 2020-02-30: Show domain prefixes (via Vlad Bokov), partially apply patch by
+#   rlaager.
 # 2020-03-05: Patch by rlaager (allocation_classes, ZoL -> openzfs).
+# 2020-10-07: Rework by gmelikov for openzfs-docs
 
-from sys import argv
-
-if len(argv) == 1:
-    path = '.'
-elif len(argv) != 2:
-    print('Usage:', argv[0], 'path')
-    exit(1)
-else:
-    path = argv[1]
+import logging
+import sys
 
 from collections import defaultdict
 from urllib.request import urlopen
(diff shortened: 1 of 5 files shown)


scripts/compatibility_matrix.py
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/env python3
+
+# License: CC0 https://creativecommons.org/share-your-work/public-domain/cc0/
+
+# A messy script that figures out ZFS features. It's very messy, sorry. I am
```
rlaager, line 5: I would remove this bit and probably the bit about man pages. (If you keep that bit, let's remove the "lazy" bit.)

scripts/compatibility_matrix.py
```diff
@@ -0,0 +1,252 @@
...
+# not responsible if this script eats your laundry.
+#
+# This uses manpages, because I'm lazy. If your manpages are wrong, you have a
+# bug.
+#
+# If the script is wrong, or could be improved, feel free to contact me on
```
rlaager, line 11: This piece needs to go. We do not want to direct complaints to the original author after we've merged this in, made a pile of changes, and he is not in a position to commit fixes.

scripts/compatibility_matrix.py
```diff
@@ -0,0 +1,252 @@
...
+# freenode, and tell me why it's wrong. My nick is zgrep.
+#
+# 2018-07-05: Created.
+# ????-??-??: Many things happened.
+# 2020-02-23: Applied patch by rlaager.
+# 2020-02-30: Show domain prefixes (via Vlad Bokov), partially apply patch by
+#   rlaager.
+# 2020-03-05: Patch by rlaager (allocation_classes, ZoL -> openzfs).
+# 2020-10-07: Rework by gmelikov for openzfs-docs
```
rlaager, line 20: I'm not sure that the history is particularly critical once this is merged in. We'll have the git history.

scripts/compatibility_matrix.py
```diff
@@ -0,0 +1,252 @@
...
+        except Exception:
+            tags = []
+    for ver in tags:
+        sources[ver] = ('https://raw.githubusercontent.com/openzfsonosx/zfs/'
+                        'zfs-{}/man/man5/zpool-features.5'.format(ver))
+    return sources
+
+
+def freebsd_pre_openzfs():
+    # TODO(gmelikov): add FreeBSD HEAD (OpenZFS version)?
```
rlaager, line 79: I think this is what you're getting at, but if not... FreeBSD is still merging the OpenZFS code into their repository, right? So there could be some lag between OpenZFS upstream and the FreeBSD tree? I believe the goal is that there are no differences (certainly at a feature level), but that doesn't mean there wouldn't be delay. So it may still make sense to get the features from their version, to account for that lag.

scripts/compatibility_matrix.py
```diff
@@ -0,0 +1,252 @@
...
+            elif line.startswith('READ\\-ONLY COMPATIBLE'):
+                readonly[guid] = (line.split()[-1] == 'yes')
+
+os_sources = sources.copy()
+os_sources.pop(openzfs_key)
+header = list(sorted(os_sources.keys()))
+header.insert(0, openzfs_key)
+header = list(zip(header, (sorted(sources[name],
+              key=lambda x: regex(r'[^0-9]', '', x) or x) for name in header)))
+# TODO(gmelikov): add Sortix?
```
rlaager, line 196: No. From what I can tell, Sortix is a toy operating system: https://sortix.org

rlaager (changes requested):
Overall, this looks good to me and I support the idea of merging this into the official documentation. I've made a few comments.

scripts/compatibility_matrix.py
gmelikov: Cleaned header info totally, add basic description.

scripts/compatibility_matrix.py
gmelikov: You've got it right, clarified it!

scripts/compatibility_matrix.py
gmelikov: It was in original script as dump "not support anything", and after your comment I've removed even this todo.

rlaager (approved):

ghfields:
OpenZFS 2.0 code on FreeBSD doesn't support edonr.  https://github.com/openzfs/zfs/pull/9901 I don't know if there is a plan to fix it.  That would make the current table a half-truth since it combines Linux and Freebsd.  Not sure if/how you would like to address this.

gmelikov:
@ghfields good catch! I'll try to work on it further then and parse such exclusions.

behlendorf (approved):
Looks good.  I think we should go ahead and merge this to make the compatibility information more easily available, then update it handle any special cases.

gmelikov:
@behlendorf I've merged it as is for now then and added a todo into issue.

PR #77, opened 2020-11-22
Remove dRAID Howto

behlendorf:
While still broadly correct the explanation of dRAID in the Howto
page no longer matches with the final integrated implementation
of dRAID.  To head-off potential confusion remove the page until
an accurate updated version is written.


docs/Basic Concepts/dRAID Howto.rst
```diff
@@ -1,417 +0,0 @@
-dRAID Howto
-===========
-
-.. note::
```
gmelikov, line 4: @behlendorf what do you think if we just update this note for now? Otherwise I see two cons:
- this page is indexed in search engines and may have links referred on it, at least we may leave page but remove all text but info about future content so people can get something instead of 404.
- deleted page nearly equals to not-to-be-updated-ever page for me (or somebody has plans on it already?)

behlendorf: Fair points.  Let me see about updating the existing page to accurately describe dRAID as it was merged.

behlendorf:
Closing.  I've opened #78 which instead updates the documentation.

PR #79, opened 2020-11-25
Include Debian security and update repos

gibmat:
- The instructions for setting up a ZFS root on Debian systems do not configure the security or update repos. If the reader doesn't realize this, they will be missing important updates between minor releases.
- Also updated the pip3 install command to use the correct package name for all Debian/Ubuntu systems.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,7 +15,7 @@ Install Prerequisites
 The dependencies are available via pip::
 
    # For Debian based distros
-   sudo apt install pip3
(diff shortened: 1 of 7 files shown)


gibmat:
Pinging @rlaager as requested in the documentation.

gmelikov (approved):

rlaager (approved):

rlaager:
Good catches. Thanks!

PR #81, opened 2020-11-26
Update RHEL/CentOS install documentation

behlendorf:
Refresh the documentation to more clear and to account for changes
for made in the recent RHEL/CentOS releases.

- Replaced most ZFS on Linux references with OpenZFS.

- Condensed the list of repositories and made it clear which repos
  are updated with new packages and which had been archived.

- Added separate instructions for RHEL/CentOS 6,7 and for RHEL/CentOS
  8 and newer since the package manager was changes from yum to dnf.

- Recommend using `yum-config-manager` or `dnf config-manager` rather
  than manually modifing the zfs.repo file.

- Converted the important notices to "notes" and reworded them.  The
  exact commands in the documention were also for much older versions
  of ZFS and were removed entirely.  This should be less of an issue
  with more modern releases.

- Add a comment explaining the modules are automatically loaded
  when a pool is detected, but they can always be loaded at boot
  time by creating a /etc/modules-load.d/zfs.conf file.

### DIFF
--- a/docs/Getting Started/RHEL and CentOS.rst
+++ b/docs/Getting Started/RHEL and CentOS.rst
@@ -1,53 +1,38 @@
 RHEL and CentOS
 ===============
 
-`kABI-tracking
-kmod <http://elrepoproject.blogspot.com/2016/02/kabi-tracking-kmod-packages.html>`__
-or
-`DKMS <https://en.wikipedia.org/wiki/Dynamic_Kernel_Module_Support>`__
-style packages are provided for RHEL / CentOS based distributions from
-the official zfsonlinux.org repository. These packages track the
-official ZFS on Linux tags and are updated as new versions are released.
-Packages are available for the following configurations:
+`DKMS`_ or `kABI-tracking kmod`_ style packages are provided for RHEL and
+CentOS based distributions from the OpenZFS repository. These packages are
+updated as new versions are released. Only the current repository for each
+major release is updated with new packages. Packages are available for the
+following configurations:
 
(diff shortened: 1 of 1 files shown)


docs/Getting Started/RHEL and CentOS.rst
```diff
@@ -1,173 +1,165 @@
...
-signing key. All official ZFS on Linux packages are signed using this
-key, and by default yum will verify a package's signature before
-allowing it be to installed. Users are strongly encouraged to verify the
-authenticity of the ZFS on Linux public key using the fingerprint listed
-here.
+To simplify installation a *zfs-release* package is provided which includes
+a zfs.repo configuration file and public signing key. All official OpenZFS
+packages are signed using this key, and by default yum or dnf will verify a
+package's signature before allowing it be to installed. Users are strongly
+encouraged to verify the authenticity of the ZFS on Linux public key using
```
gmelikov, line 17:
```suggestion
encouraged to verify the authenticity of the OpenZFS public key using
```

gmelikov: Maybe I'm wrong and it was intentional.

gmelikov (approved):

docs/Getting Started/RHEL and CentOS.rst
behlendorf: This was actually intentional since the name on the key itself is still "ZFS on Linux" and I wanted to avoid any confusion.

PR #88, opened 2020-12-09
use apt instead of dpkg for os-prober removing

Yvan-Masson:
Hi,

In general `apt` is the preferred way to interact with packages, and I think in this case there is no need/advantage to use `dpkg`.

Regards,
Yvan

### DIFF
--- a/docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
@@ -650,7 +650,7 @@ Step 4: System Configuration
 
 #. Optional: Remove os-prober::
 
-     dpkg --purge os-prober
+     apt --purge os-prober
 
    This avoids error messages from `update-grub`.  `os-prober` is only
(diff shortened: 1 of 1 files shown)


rlaager:
Seems reasonable. Can you update the Ubuntu 20.04 version too?

In general, the older versions for both distros can be changed or not. In this case, if they already use apt (instead of apt-get), it is fine to update them too.

Yvan-Masson:
OK, I will take a look on Ubuntu and older versions to see if the same change applies. But I don't know if I will manage to do this in one pull request :-) (I am not at ease with Github).

rlaager:
I redid this (with credit to you) as an edit for all of the files.

In general, what you would do is:
1. Edit all the files.
2. `git commit -a` to commit all changes at once. Or `git add` each file and then `git commit`.

If you're using the GitHub web interface to edit, I have no idea if it's even possible to make it one commit or one PR.

Anyway, thanks for the patch!

Yvan-Masson:
Thanks for the credits and for the explanation!

PR #93, opened 2020-12-18
Fix removing the os-prober

suda:
Thank you so much for this document! While following the instructions I found one line has a typo and I hope this PR can help :)


rlaager:
Thanks for the fix! I applied this to all three affected files. A recent change broke that. BTW, the double-colon is correct. That's how .rst does code blocks. It's weird, but that's how it works.

PR #96, opened 2020-12-23
Update OpenZFS exception list

jumbi77:
Although openzfs/zfs is the new default repo and the OpenZFS patch exceptions are obsolete, the list should be as accurate as possible.


behlendorf (approved):
Thanks!

PR #105, opened 2021-01-14
Add support for Alpine Linux

ne9z:
Articles are directly viewable at [here](https://github.com/ne9z/openzfs-docs/tree/alpine-support/docs/Getting%20Started/Alpine%20Linux).

This pull request adds instructions for installing ZFS package and Root on ZFS installation for Alpine Linux.

@rlaager

### DIFF
--- a/docs/Getting Started/Alpine Linux/Alpine Linux Root on ZFS.rst
+++ b/docs/Getting Started/Alpine Linux/Alpine Linux Root on ZFS.rst
@@ -1,0 +1,708 @@
+.. highlight:: sh
+
+Alpine Linux Root on ZFS
+========================
+
+.. contents:: Table of Contents
+   :local:
+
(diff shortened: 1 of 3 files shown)


ne9z:
ZFS is present in the official repo, so just using ZFS on other disks is straightforward.

However, currently there're some limitations regarding Root on ZFS:
- Bugs in init script prevents decrypting root pool. A solution is included.
- Initramfs does not support `/dev/disk/by-*` and the maintainer so far declined to include `eudev` in initramfs:
> I don't think its a good idea to add support for udev in initramfs. We should instead figure out how to create the /dev/disk/by-id/* from nlplug-findfs.c.

Therefore multi-disk setup is discouraged.

ne9z:
Closing. Alpine initramfs is too buggy.

PR #106, opened 2021-01-15
Add support for Void Linux

ne9z:
Articles are directly viewable at [here](https://github.com/ne9z/openzfs-docs/tree/void-support/docs/Getting%20Started/Void%20Linux).

### DIFF
--- a/docs/Getting Started/Void Linux/Void Linux Root on ZFS.rst
+++ b/docs/Getting Started/Void Linux/Void Linux Root on ZFS.rst
@@ -1,0 +1,773 @@
+.. highlight:: sh
+
+Void Linux Root on ZFS
+=======================
+
+.. contents:: Table of Contents
+   :local:
+
+Overview
+--------
+
+Caution
+~~~~~~~
+
+- This guide uses entire physical disks.
+- Multiple systems on one disk is not supported.
+- Target disk will be wiped. Back up your data before continuing.
+- The target system, virtual or physical, must have at least 4GB RAM,
+  or the DKMS module will fail to build.
+
+Support
+~~~~~~~
+
+If you need help, reach out to the community using the :ref:`mailing_lists` or IRC at
+`#zfsonlinux <irc://irc.freenode.net/#zfsonlinux>`__ on `freenode
+<https://freenode.net/>`__. If you have a bug report or feature request
+related to this HOWTO, please `file a new issue and mention @ne9z
+<https://github.com/openzfs/openzfs-docs/issues/new?body=@ne9z,%20I%20have%20the%20following%20issue%20with%20the%20Void%20Linux%20Root%20on%20ZFS%20HOWTO:>`__.
+
+Contributing
+~~~~~~~~~~~~
+
+#. Fork and clone: https://github.com/openzfs/openzfs-docs
+
+#. Install the tools::
+
+    sudo xbps-install python3-pip
+
+    pip3 install -r docs/requirements.txt
+
+    # Add ~/.local/bin to your $PATH, e.g. by adding this to ~/.bashrc:
+    PATH=$HOME/.local/bin:$PATH
+
+#. Make your changes.
+
+#. Test::
+
+    cd docs
+    make html
+    sensible-browser _build/html/index.html
+
+#. ``git commit --signoff`` to a branch, ``git push``, and create a pull
+   request. Mention @rlaager.
+
+Encryption
+~~~~~~~~~~
+
+This guide supports optional ZFS native encryption on the root pool.
+
+Boot pool, where ``/boot`` is located, is not encrypted.
+
+ZFS native encryption does not encrypt metadata. All datasets properties
+are available immediately upon importing, without the key.
+
+Preinstallation
+----------------
+
+Prepare the Live Environment
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Visit `Void Linux mirror <https://alpha.de.repo.voidlinux.org/live/>`__,
+   choose the most recent directory,
+   download
+   ``void-live-x86_64-musl-DATE.iso`` or
+   ``void-live-x86_64-DATE.iso``
+   and write it to a USB drive or an optical disc.
+
+   Do not download from ``current`` or the version listed on the official
+   `Download page <https://voidlinux.org/download/>`__.
+
+   They contain outdated kernel and require
+   an update to install headers, which is not doable in live environment.
+
+#. Boot the target computer from the prepared live medium.
+
+#. Connect to the internet. 
+   If the target computer aquires IP address with DHCP, 
+   no further steps need to be taken. 
+   Otherwise, refer to 
+   `Void Linux Handbook <https://docs.voidlinux.org/config/network/index.html>`__ 
+
+#. Login as root. Instructions are available in MOTD.
+
+
+#. Start SSH server.
+
+    - Permit root login with password::
+
+       echo PermitRootLogin yes >> /etc/ssh/sshd_config
+
+    - Restart SSH server::
+
+       sv restart sshd
+
+    - Find the IP address of the target computer::
+
+       ip -4 address show scope global
+
+    - On another computer, connect to the target computer with::
+
+       ssh root@192.168.1.10
+
+#. Enter a bash shell::
+
+     bash
+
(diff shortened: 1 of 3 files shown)


ne9z:
Withdrawn until [linux-headers package is added to live image](https://github.com/void-linux/void-mklive/pull/179).

zdykstra:
As a Void user and one of the authors of [ZFSBootMenu](https://zfsbootmenu.org/), I don't think this guide represents a meaningful picture of where booting root-on-ZFS under Void stands right now. I am, of course, biased - but the install guides available at https://github.com/zbm-dev/zfsbootmenu/wiki are both far more concise and quite tested. 

There are a substantial number of extraneous steps listed in this PR that should be removed both because they are highly specific to local installation needs and because they greatly complicate what should be a fairly simple installation process.

ne9z:
@zdykstra: In this PR I'm mostly trying to duplicate the guides for Debian and Ubuntu.

I had a look at your guide, but I found that it's somewhat insecure and has important steps omitted.

In [Void Linux Single disk UEFI](https://github.com/zbm-dev/zfsbootmenu/wiki/Void-Linux---Single-disk-UEFI)
- Key is embedded in initramfs, which in turn resides on the unencrypted EFI partition.
- Initramfs is not protected by ZFS. With separate boot pool, root filesystem and `/boot` can be rolled back together with only an option in GRUB menu. This is implemented in bieaz, a BE manager I wrote.
- Requires glibc
- Requires 3rd-party hrmpf image. It would be great if it's official like Alpine Extended
- Important persistent data such as `/var/lib/docker` are in the same dataset as the root filesystem

> Quite tested

I assure you, except the `linux-headers` part in official live image, this PR is also quite tested.

> There are a substantial number of extraneous steps listed in this PR

Feel free to make comments on this issue.

zdykstra:
> @zdykstra: In this PR I'm mostly trying to duplicate the guides for Debian and Ubuntu.
> 
> I had a look at your guide, but I found that it's somewhat insecure and has important steps omitted.
> 
> In [Void Linux Single disk UEFI](https://github.com/zbm-dev/zfsbootmenu/wiki/Void-Linux---Single-disk-UEFI)
> 
>     * Key is embedded in initramfs, which in turn resides on the unencrypted EFI partition.

No, you've misread the guide. The key is embedded in the initramfs that is on an encrypted ZFS dataset. ZFSBootMenu uses an intermediary initramfs stored on your ESP; it does not have any encryption keys in it.
> 
>     * Initramfs is not protected by ZFS. With separate boot pool, root filesystem and `/boot` can be rolled back together with only an option in GRUB menu. This is implemented in bieaz, a BE manager I wrote.

Again, you've misunderstood what ZFSBootMenu is. Your actual OS kernels and initramfs files are stored on /boot, which is backed by ZFS. 

> 
>     * Requires glibc
It doesn't require glibc, I just set that in the documentation because it's the most useful out-of-the-box for people.

> 
>     * Requires 3rd-party hrmpf image. It would be great if it's official like Alpine Extended
> 
>     * Important persistent data such as `/var/log` are in the same dataset as the root filesystem
That's up for debate. Logs are, in my experience, tied specifically to the state of a boot environment. As such, they should be associated with it.
> 
> 
> > Quite tested
> 
> I assure you, except the `linux-headers` part in official live image, this PR is also quite tested.
> 
> > There are a substantial number of extraneous steps listed in this PR
> 
> Feel free to make comments on this issue.

Again, your guide does not represent what is considered an optimal or even good-enough Void system.

ne9z:
@zdykstra: Great to hear that zfsbootmenu does not require a separate pool to work and simplifies the installation.

So really, why don't submit a PR to update the Debian/Ubuntu guide to eliminate separate boot pool and just use the boot menu in external initramfs approach?

It would also be helpful if your Void and Debian guide are listed as official OpenZFS docs.

ahesford:
If you want logs to be decoupled from a BE, it is trivial to use a separate dataset for that purpose and still enjoy the benefits of a consistent BE that ZFSBootMenu uses.

Using a separate boot pool is a bad practice because snapshots of boot and root cannot be guaranteed to be consistent, requiring careful coordination of two independent pools and file systems. (This is the motivation for Ubuntu's `zsys` hack.) Furthermore, GRUB can't really provide meaningful rollback facilities, instead requiring pre-made clones of snapshots and a pre-configured menu to choose them.

ZFSBootMenu makes consistency a non-issue by directly inspecting BEs, each *containing their own `/boot`*, allowing you to work from one set of snapshots. You can list and investigate past snapshots *ex post facto* and decide to clone (or make a completely independent copy) *on demand* and boot it.

ahesford:
Furthermore, because ZFSBootMenu supports completely atomic boot environments, it extends the concept of "boot environments" beyond "clones of a single system" to "any number of arbitrary Linux distributions". Thus, alongside my regular Void BE, I have a Void musl BE and, for testing, an Arch BE. With a few arrow keys, I can boot any of these completely unrelated operating systems on the same hardware.

ne9z:
@zdykstra, @ahesford: Great stuff! This dramatically simplifies the installation and does not need a separate bpool.

I apologize for misreading the guide and not really understanding zbm previously.

As I understand, zbm is a ZFS-enabled initramfs that allow user to manipulate the filesystem, including operations such as creating boot environments.

One drawback is that, zbm initramfs then becomes critical for booting the system. Therefore it might be a sensible approach to store zbm on a zpool instead of a ext4 partition and let GRUB to load zbm.

zbm also must be regenerated when a new zpool feature becomes available.

I'd argue against manage and discover other distros with zbm. That is really the job of the bootloader.

In your Void/Arch example, a separate zbm for each distro might be a better approach. Just add items to GRUB menu:
```
Arch Linux zfsbootmenu
Void Linux zfsbootmenu
Microsoft Windows
```

ne9z:
I suspect that making the system completely dependent on zfsbootmenu to boot is the reason why the inferior b/rpool approach is adopted by Debian/Ubuntu guide instead.

ahesford:
ZFSBootMenu **is** the boot loader. The point is to replace GRUB, which does a poor job supporting ZFS and gives you none of the flexibility that ZFS offers *when you need it*, which is before the system is booted.

You can either completely depend on GRUB or completely depend on ZFSBootMenu. You're always beholden to the boot loader you choose.

ZFSBootMenu is meant to work with dracut, but this has no compatibility ramifications. The ZBM initramfs is independent of the operating system and is used only to provide the pre-boot environment. After you select a BE to boot, it jumps into the kernel and uses the initramfs image native to that environment. That image can be generated by dracut, mkinitcpio, initramfs-tools or whatever method your distribution of choice uses. Users need to install dracut to produce ZBM images, but they can otherwise ignore the existence of dracut on their systems.

rlaager:
> So really, why don't submit a PR to update the Debian/Ubuntu guide to eliminate separate boot pool and just use the boot menu in external initramfs approach?

Please don't waste your time on that. The intention with both HOWTOs is that the resulting systems should be as close as possible to a default, non-ZFS install as possible. Accordingly, it is specifically a goal to use GRUB. FWIW, the Ubuntu team agreed with this approach when building their installer support.

I suggest a similar approach for other distributions. The result should be as close to possible as "stock". If people want to make further customizations, more power to them.

ne9z:
@rlaager: No, I haven't devoted anytime to adopt the three guides, -- Arch, Void, Alpine --, to ZFSBootMenu.

I'm still supporting separate boot pool + GRUB approach.

By the way, how is the Arch guide going? Are there any changes I should make?

ahesford:
Debian and Ubuntu can do whatever they think is best for their users. I agree that OpenZFS documentation should reflect what upstream considers best practices. As a member of both Void Linux and ZFSBootMenu, I do not believe this is an advisable approach for Void. There is a long-standing [open PR](https://github.com/void-linux/void-docs/pull/206/files) to include root-on-ZFS documentation for Void that recommends using GRUB to find `/boot` on the same pool as the root, but this has not been merged precisely because there is no strong support for this approach and there is no consensus about a suitable alternative. I would prefer to see OpenZFS refrain from sanctioning an approach for Void until the Void team has arrived at a consensus on a recommended procedure.

ne9z:
@ahesford: this PR is already closed and at least myself will not put any effort towards supporting this Void on ZFS guide.

Without official support from Void, no guide will materialize here.

PR #109, opened 2021-01-20
Corrected partition number in grub config for efi

rakor:
In fstab there should be the right partition for /boot/efi.


rlaager (approved):

rlaager:
Thanks!

PR #113, opened 2021-01-29
Use archive repo for rolling kernel releases

ne9z:
Arch Linux and its derivatives are special in that the kernel is rapidly updated whenever there's an update available upstream. This leads to compatibility problems with OpenZFS. Complicating things further is the support for multiple kernel variants.

Fortunately there's a solution for this: Arch Archive Repo, which retains old packages for up to 2 years, see [wiki](https://wiki.archlinux.org/index.php/Arch_Linux_Archive). We can then easily downgrade the kernel to any desired version.

This PR adds instructions for performing such a downgrade.

@rlaager

### DIFF
--- a/docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
+++ b/docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
@@ -22,16 +22,6 @@ Caution
   will not work on 4Kn with legacy (BIOS) booting.
   <http://savannah.gnu.org/bugs/?46700>`__
 
-.. note::
-
-    Due to the release cycle of OpenZFS and the rapid adoption of new kernels
-    it may happen that you won’t be able to
-    build DKMS packages for the most recent kernel update.
-    If the `latest OpenZFS release <https://github.com/openzfs/zfs/releases/latest>`__
-    does not yet support the installed kernel,
-    `use an older live image <https://mirrors.dotsrc.org/archlinux/iso/>`__
-    before installation.
-
 Support
 ~~~~~~~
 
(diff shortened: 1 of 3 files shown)


rlaager:
It sounds like there are two issues here: 1) needing a Live CD that has a kernel old enough to be compatible with OpenZFS, and 2) preventing the installed system from upgrading to a kernel that will break it.

For #1, if it were me, I'd link to a specific version of the Live CD, and just be clear that the user needs to use that version. There's very little gain for using a new Live CD, since it's only used for the install anyway, right? So it doesn't seem all that necessary to have "use the latest, but check to see if it's okay, and if not use the old one" like you have here. I'd just say, "use this one".

For #2, is there any way to prevent the system from upgrading to an incompatible kernel? For example, is there any sort of packaging "pinning" method in Arch? If not, I suppose all you can do is warn the user loudly in the HOWTO.

ne9z:
Arch Linux only retains live images for 3 months. So there's a need to update the link when the old image got deleted. Nevertheless, I think this is a sensible approach. I've added fixed live image links to the guide and a dedicated section for checking live image compat.

For DKMS package, a step for holding kernel and header packages has been added.

This is not necessary for prebuilt packages as they hard-depend on a compatible kernel version, preventing kernel update if a corresponding prebuilt package is not available.

rlaager:
I'm not planning to review commits to this HOWTO in depth. So just let me know when you think this PR is ready to merge.

ne9z:
I think it's ready to merge.

ne9z:
Don't merge till 2nd Feb so I can update the LiveCD link.

ne9z:
@rlaager I think it's ready now, please merge.

PR #117, opened 2021-02-06
Handle glibc mismatch

ne9z:
`zfs-dkms` might depend on an yet unavailable glibc version when using a newer kernel.

This commit advises the user to downgrade kernel in this scenario.

### DIFF
--- a/docs/Getting Started/Arch Linux/index.rst
+++ b/docs/Getting Started/Arch Linux/index.rst
@@ -240,6 +240,53 @@ If compatible, update kernel with::
 Do not update if the kernel is not compatible
 with OpenZFS.
 
+glibc version mismatch
+^^^^^^^^^^^^^^^^^^^^^^
+As of Feb 6th, 2021, latest glibc release is ``2.33``.
+glibc in Arch Linux repo is ``2.32``.
+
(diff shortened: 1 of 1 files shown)


ne9z:
@rlaager

rlaager:
> When updating ``linux-lts (5.4.94-1 -> 5.4.95-1)``, ``zfs-dkms`` will depend on the unavailable glibc ``2.33``

Why is this happening? Why would a kernel upgrade make some other package depend on a newer version of glibc that is not yet packaged in the distro?

ne9z:
@rlaager: Somehow `linux-lts-headers` depends on that glibc that's not yet packaged. See [this reddit post](https://ol.reddit.com/r/archlinux/comments/ldg4oo/arch_linux_linuxlts_kernel_borked/).

ne9z:
Likely this is the packager's fault.

rlaager:
Right, that makes more sense. You want to squash these together?

BTW, can you start using `--signoff` on your commits, so they get a `Signed-off-by` header, please? We try to use those here.

ne9z:
@rlaager: Squashed and signed off.

PR #119, opened 2021-02-07
disable automounting on live environment

apfelchips:
751: "mkdosfs -F 32 -s 1 -n EFI ${DISK}-part1" fails because some partitions are automounted --> disable automouting

### DIFF
--- a/docs/Getting Started/Ubuntu/Ubuntu 20.04 Root on ZFS.rst
+++ b/docs/Getting Started/Ubuntu/Ubuntu 20.04 Root on ZFS.rst
@@ -253,6 +253,9 @@ Step 1: Prepare The Install Environment
    ``ip addr show scope global | grep inet``. Then, from your main machine,
    connect with ``ssh ubuntu@IP``.
 
+#. Disable Automounting
+     gsettings set org.gnome.desktop.media-handling automount false
+
 #. Become root::
 
      sudo -i


rlaager:
Is this a disk where that filesystem already existed (e.g. you are reinstalling on top of an existing install)?



I can edit your commit as I merge, but in case you want to:

The steps are using sentence caps, not title case, so "Automounting" should be lowercase. In terms of formatting, you need a double colon and a blank line before the code block. That is, instead of:
```
#. Disable Automounting
￼     gsettings set org.gnome.desktop.media-handling automount false
```
you need:
```
#. Disable automounting::

￼     gsettings set org.gnome.desktop.media-handling automount false

   ...
```

and we should probably add some explanatory text so people (including future me) will know what this is for.

So maybe something like this:
```
#. Disable automounting:

   If the disk has been used before (with partitions at the same offsets),
   previous filesystems (e.g. the ESP) will automount if not disabled::

￼     gsettings set org.gnome.desktop.media-handling automount false
```

Commit with `git commit --signoff` so it gets the `Signed-off-by` header. And if you're taking my explanatory text, then if we're being pedantic, it should probably have a `Co-authored-by: Richard Laager <rlaager@wiktel.com>` too.

To edit an existing PR, you would make the changes, then `git commit --amend` and then force push with `git push -f apfelchips patch-1`.

Again, this is just if you want to make the edits. If you did this in the GitHub editor (which I think you might have given the `patch-1`) and you'd rather I make the edits, that's fine.

apfelchips:
Actually I just wanted to add this little paragraph and thought the github editor would do all the things needed, sadly iseemingly it doesn't.  Would be great if you could do the edit as you've everything set up. Thanks.

I've wanted to replace my freenas system with this, but my pools have some newer features enabled, so that didn't work. Will there be a backport of OpenZFS2.0 to 20.04 LTS or will we have to wait for 21.04?

PR #121, opened 2021-02-15
Arch Linux: Remove alternative solutions

ne9z:

The previous approach to installation is: try this easy one first, and if it fails, try the hard one.
This commit streamlines the installation experience by minimizing references to such procedures.

This commit also removes verbose comments from Root on ZFS installation guides.

As the upstream has fixed glibc, its section is also removed.

@rlaager


ne9z:
@rlaager: Merged formatting fixes.

There's still one warning for Debian guide:
```
docs/Getting Started/Debian/Debian Buster Root on ZFS.rst:134: WARNING: Literal block expected; none found.
```

PR #124, opened 2021-02-20
Arch Linux: Multiple fixes for Root on ZFS guides

ne9z:
This commit includes multiple fixes for Arch/Artix Linux Root on ZFS guides.

- Separate archzfs repo mirrorlist from /etc/pacman.conf
- Include `nano` editor when editing text files
- Fix Linux package variant detection command
- Arch only: expand live root file system to 2GB
- Update glibc when installing zfs-dkms
- Artix-only: specify elogind-openrc
- Artix-only: fix repo name 'core' -> 'system'


ne9z:
@rlaager

Issue #107, opened 2021-01-16
Ubuntu 20.04 Root on ZFS - No mirrored ESP Partition


JohannesNeu:
@rlaager, I have the following issue with the Ubuntu 20.04 Root on ZFS HOWTO:

Hi,
I am at step 6.5 "dpkg-reconfigure grub-efi-amd64". I can run the command and questions will be asked, but there is no possibility to choose the EFI partitions from a list. I have a mirror setup with two SSDs and each has an EFI partition, a BOOT partition and a ROOT partition. 
These are the questions I am asked:
![grafik](https://user-images.githubusercontent.com/39328632/104819961-b350f080-5831-11eb-81ea-0d59a0e991b8.png)
![grafik](https://user-images.githubusercontent.com/39328632/104819966-c19f0c80-5831-11eb-854b-281de2b9960a.png)
![grafik](https://user-images.githubusercontent.com/39328632/104819970-cebbfb80-5831-11eb-8573-1013af796d9c.png)

And this is the result:
![grafik](https://user-images.githubusercontent.com/39328632/104820003-19d60e80-5832-11eb-8103-ad1e276e62c5.png)

Is it possible that this is connected to my /etc/fstab setup?
![grafik](https://user-images.githubusercontent.com/39328632/104820018-4558f900-5832-11eb-87d3-93d22d5a5298.png)

Thanks for your help.

rlaager:
I'm not sure. I've certainly seen that be an issue before the system reboots (which is why I save that step until after the reboot). I'm not sure about the /etc/fstab entry, but you could try making it `/dev/disk/by-uuid/B948-62E2 /boot/efi vfat defaults 0 0` instead. If that works, let me know and I'll have to think about removing the recently added (c00aa93389260573349303e95cb3e6e3cb805b09) automount idea I guess?

Are you running into this issue:
https://github.com/openzfs/openzfs-docs/issues/95

JohannesNeu:
@rlaager Thanks for the link, I totally missed that issue.
I checked: and it seems I also have grubenv:


`ich@host:/boot/grub$ ls -la`
`total 1644`
`drwxr-xr-x 4 root root       7 Jan 17 09:37 .`
`drwxr-xr-x 4 root root      12 Jan 16 16:02 ..`
`drwxr-xr-x 2 root root       3 Jan 16 14:00 fonts`
`-r--r--r-- 1 root root   13791 Jan 17 09:37 grub.cfg`
[...]

PR #125, opened 2021-02-22
Arch Linux: Multiple fixes

ne9z:
- Update add archzfs repo instructions
- Use META for determining compatible Linux version

@rlagger


docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
```diff
@@ -127,24 +127,18 @@ Prepare the Live Environment
 
 #. Import keys of archzfs repository::
 
-    curl -O https://archzfs.com/archzfs.gpg
-    pacman-key -a archzfs.gpg
-    pacman-key --lsign-key DDF7DB817396A49B2A2723F7403BD972F75D9D76
```
rlaager, line 132: Personally, I think it's better to keep the key ID hardcoded here. I suppose if someone's typing it in, it's harder, but if they're copying-and-pasting, it's the same. And it's not clear to me who controls git.io/JtHEx. And same for the couple places below.

rlaager (changes requested):

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
ne9z: Both key-id link and mirrorlist link is maintained by myself, at [this repo](https://github.com/ne9z/mirrorlist-archzfs).

Using a link instead of a hard-coded value is not only for convenience of the user, but also for maintaining the guide.

This key-id [has been changed before](https://github.com/archzfs/archzfs/issues/262). It might be easy for me to replace 5 occurrences in these guides, but not so much for replacing its derivatives. If someone copy-and-pasted the code to their own blog, script or something, it's up to them to update the key-id. Just a quick search on [Google](https://www.google.com/search?q=%22DDF7DB817396A49B2A2723F7403BD972F75D9D76%22) turned up many instances.

ne9z: For clarification:
- git.io/JtHEx points to https://github.com/ne9z/mirrorlist-archzfs/raw/master/key-id
- git.io/JtHEh points to https://github.com/ne9z/mirrorlist-archzfs/raw/master/mirrorlist-archzfs

ne9z:
@rlaager: Or maybe we could host mirrorlist and keyfile right in this repo under `docs/Getting Started/Arch Linux/{archzfs-key-id,archzfs-mirrorlist}`? This way the trust is placed on this repo.

ne9z:
The above commit moves mirrorlist and key-id to this repo.

Links used:
- https://git.io/JtQpl to https://github.com/openzfs/openzfs-docs/raw/master/docs/Getting%20Started/Arch%20Linux/archzfs/key-id
- https://git.io/JtQp4 to https://github.com/openzfs/openzfs-docs/raw/master/docs/Getting%20Started/Arch%20Linux/archzfs/mirrorlist-archzfs

ne9z:
@rlaager: The above comment adds systemd.automount for ESP.

Regarding #107, I noticed that on Arch, systemd 247.3-1, when using 
```
x-systemd.idle-timeout=1min,x-systemd.automount,noauto
```
, ESP appears twice in `/proc/mounts`:
```
systemd-1 /boot/efi autofs rw,relatime,fd=47,pgrp=1,timeout=60,minproto=5,maxproto=5,direct,pipe_ino=10567 0 0
r
/dev/vda1 /boot/efi vfat rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,utf8,errors=remount-ro 0 0
```

Previously if ESP fails to mount at boot, systemd will drop into emergency shell, which is not desirable. This allows ESP mounting failure.

ne9z:
Defaults to encryption with key file. Also added instructions to use password and remote SSH unlock.

ne9z:
@rlaager: Tests against Arch/Artix completed. Ready for merge.

rlaager (changes requested):
I don't like this bit about the unprotected key file. Changing the password later does not change the master key. Granted, it does change the wrapping key, and it's unlikely that someone would compromise the machine in between the reboot. But why aren't you just using a passphrase initially?

rlaager:
> The above commit moves mirrorlist and key-id to this repo.
> 
> Links used:
> 
> * https://git.io/JtQpl to https://github.com/openzfs/openzfs-docs/raw/master/docs/Getting%20Started/Arch%20Linux/archzfs/key-id
> * https://git.io/JtQp4 to https://github.com/openzfs/openzfs-docs/raw/master/docs/Getting%20Started/Arch%20Linux/archzfs/mirrorlist-archzfs

I'm not seeing this change. It still references git.io from what I can see.

ne9z:
@rlaager:
> I'm not seeing this change. It still references git.io from what I can see.
I mean the git.io links are pointing to them. You can see it for your self with `curl -v https://git.io/JtQpl`, they will be available after merging.

ne9z:
@rlaager 
> I don't like this a bit about unprotected keyfile

This is inspired by [this ubiquity issue](https://bugs.launchpad.net/ubuntu/+source/ubiquity/+bug/1857398).
> We should therefore always install with encryption enabled on zfs systems, with a non-secret key by default, and allow the user to use 'zfs change-key -o keylocation=prompt' after install to take ownership of the encryption and upgrade the security.

rlaager:
> @rlaager
> 
> > I don't like this a bit about unprotected keyfile
> 
> This is inspired by [this ubiquity issue](https://bugs.launchpad.net/ubuntu/+source/ubiquity/+bug/1857398).
> 
> > We should therefore always install with encryption enabled on zfs systems, with a non-secret key by default, and allow the user to use 'zfs change-key -o keylocation=prompt' after install to take ownership of the encryption and upgrade the security.

That was a terrible idea, and that was explained many times on that issue.

ne9z:
@rlaager: Somehow I missed all the comments, sorry! I'm reverting this change.

ne9z:
@rlaager: I labeled default keyfile encryption as "unencrypted". Still better than nothing.

rlaager:
It's not unencrypted though. It'll come with the performance penalty of encryption. Just offer people unencrypted and encrypted (with a passphrase).

ne9z:
@rlaager: Restored to unencrypted/password encryption.

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
```diff
@@ -396,7 +405,7 @@ Create Root and Boot Pools
         -O xattr=sa \
         -O mountpoint=/ \
         -R $INST_MNT \
-        -O encryption=aes-256-gcm \
+        -O encryption=on \
```
rlaager, line 408: Is this OpenZFS 2.0 or an older version?

ne9z: No, latest 2.0.3.

ne9z: Defaults to aes-256-gcm, as you know.

rlaager: To be clear, if it's OpenZFS 2.0, then on is the same as aes-256-gcm, so "on" is good. If it's older, then you should use "aes-256-gcm" explicitly. If it's OpenZFS 2.0, can you just confirm for me.

Also, can you squash this (at least somewhat)? Then it should be good to merge.

rlaager:
[Our messages on the encryption cross, so you already answered that piece. I've marked it as resolved.]

Can you squash this (at least somewhat) and then I'll merge it?

ne9z:
@rlaager: Squashed to one commit.

PR #127, opened 2021-02-24
Arch Linux: Nest system datasets; LUKS bpool; secure boot

ne9z:
- Tell readers that encryption can not be enabled later pool-wide
- Add some warning messages for choosing a strong password
- Remove remains of default-key

@rlaager

### DIFF
--- a/docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
+++ b/docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
@@ -73,6 +73,9 @@ without the passphrase being entered at the console. Performance is
 good. As the encryption happens in ZFS, even if multiple disks (mirror
 or raidz topologies) are used, the data only has to be encrypted once.
 
+Boot pool can be optionally encrypted with LUKS, see `here <#encrypt-boot-pool-with-luks>`__.
+Encrypted boot pool can protect initrd from tempering.
+
 Preinstallation
 ----------------
 Download Arch Linux live image
@@ -291,6 +294,9 @@ Format and Partition the Target Disks
        sgdisk -n4:0:0   -t4:8308 $DISK
 
     Adjust the swap partition size to your needs.
+    If `hibernation <#hibernation>`__ is needed,
+    swap size should be same or larger than RAM.
+    Check RAM size with ``free -h``.
 
 #. Repeat the above steps for other target disks, if any.
 
@@ -322,6 +328,7 @@ Create Root and Boot Pools
 
     zpool create \
         -o ashift=12 \
+        -o autotrim=on \
         -d -o feature@async_destroy=enabled \
         -o feature@bookmarks=enabled \
         -o feature@embedded_data=enabled \
@@ -374,12 +381,12 @@ Create Root and Boot Pools
      “invalid dnode type” error. This feature does not matter for ``/boot``
      anyway.
 
-#. Create root pool:
-
-   - Unencrypted::
+#. Create root pool::
 
        zpool create \
         -o ashift=12 \
+        -o autotrim=on \
+        -R $INST_MNT \
         -O acltype=posixacl \
         -O canmount=off \
         -O compression=zstd \
@@ -388,26 +395,6 @@ Create Root and Boot Pools
         -O relatime=on \
         -O xattr=sa \
         -O mountpoint=/ \
-        -R $INST_MNT \
-        rpool_$INST_UUID \
-        ${DISK}-part3
-
-   - Encrypted::
-
-       zpool create \
-        -o ashift=12 \
-        -O acltype=posixacl \
-        -O canmount=off \
-        -O compression=zstd \
-        -O dnodesize=auto \
-        -O normalization=formD \
-        -O relatime=on \
-        -O xattr=sa \
-        -O mountpoint=/ \
-        -R $INST_MNT \
-        -O encryption=on \
-        -O keylocation=prompt \
-        -O keyformat=passphrase \
         rpool_$INST_UUID \
         ${DISK}-part3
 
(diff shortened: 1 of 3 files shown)


ne9z:
Actually I think enabling encryption on the pool dataset `rpool` is quite inflexible. The only way to get rid of the encryption or change any readonly dataset property is to destroy and rebuild the pool. A number of commands, such as `zfs rename`, [can not be used on the pool dataset](https://serverfault.com/questions/440632/zfs-rename-move-root-filesystem-into-child).

If we store everything under a child dataset, say `rpool/arch`, only enable encryption on this dataset, disable encryption, or rebuild system after password leak, is relatively simple as copying files.

This also allows multi-distro, each with their own encryption password at boot:
```
rpool/arch/...
rpool/artix/...
rpool/debian/...
EFI/arch/grubx64.efi
EFI/artix/grubx64.efi
EFI/debian/grubx64.efi
```

Maybe this is the way forward?

ne9z:
Worked out the changes. Datasets are nested under `pool/sys`.

ne9z:
Completed test run, so far so good. Ready for merge:
```
[root@zfslive ~]# zfs list -o name,encryption,encryptionroot,keylocation,keyformat
NAME                                     ENCRYPTION   ENCROOT           KEYLOCATION  KEYFORMAT
bpool_650xc9                             off          -                 none         none
bpool_650xc9/sys                         off          -                 none         none
bpool_650xc9/sys/BOOT                    off          -                 none         none
bpool_650xc9/sys/BOOT/default            off          -                 none         none
rpool_650xc9                             off          -                 none         none
rpool_650xc9/sys                         aes-256-gcm  rpool_650xc9/sys  prompt       passphrase
rpool_650xc9/sys/DATA                    aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default            aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default/home       aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default/root       aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default/srv        aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default/usr        aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default/usr/local  aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default/var        aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default/var/lib    aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default/var/log    aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default/var/spool  aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/DATA/default/var/tmp    aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/ROOT                    aes-256-gcm  rpool_650xc9/sys  none         passphrase
rpool_650xc9/sys/ROOT/default            aes-256-gcm  rpool_650xc9/sys  none         passphrase
```

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
```diff
@@ -333,14 +333,7 @@ Create Root and Boot Pools
         -o feature@large_blocks=enabled \
         -o feature@lz4_compress=enabled \
         -o feature@spacemap_histogram=enabled \
-        -O acltype=posixacl \
```
rlaager, line 336: Why move these from the top-level to the child dataset?

rlaager:
> Actually I think enabling encryption on the pool dataset `rpool` is quite inflexible. The only way to get rid of the encryption or change any readonly dataset property is to destroy and rebuild the pool. A number of commands, such as `zfs rename`, [can not be used on the pool dataset](https://serverfault.com/questions/440632/zfs-rename-move-root-filesystem-into-child).

Right, those are the downsides.

Some advantages of making the top-level dataset the encryption root are:
1. Everything can be encrypted at once: "OS" datasets like rpool/ROOT/ubuntu_UUID (and its children), OS-managed home directory datasets like rpool/USERDATA/username_UUID, and administrator-managed datasets like rpool/srv (or whatever). Obviously, some bits of that layout are zsys-specific, but the general concepts predate zsys (e.g. see the Debian Buster HOWTO).
2. This is conceptually similar to "full disk encryption", so it fits into existing designs (e.g. installers) easily.
3. It avoids the need for an intermediate dataset level (in your case, you used rpool/sys), which is: A) just nicer, B) backwards compatible with existing installs, and C) consistent between encrypted and unencrypted setups. That said, consistency can be maintained by having the additional dataset level in both encrypted and unencrypted setups, which I think is what you have done.

> If we store everything under a child dataset, say `rpool/arch`, only enable encryption on this dataset. Then disable encryption, or rebuild system after password leak, could a relatively simple prcedure.

Disabling encryption is perfectly fine. But if you want to enable encryption or change the key, there's no great way to _guarantee_ that the old data is off the disk. Granted, with SSDs and TRIM, it's pretty close. And there is a trade-off: if we want to make an absolute guarantee, then the answer is "wipe the disk and rebuild your pool", but that's not great for e.g. individual desktop users, and getting them something good might be better than nothing.

Personally, I don't think that encrypted -> unencrypted is a transition that people generally are about these days. The other direction (enabling encryption) is probably more common. But that said, I _personally_ don't see a lot of value in being able to transition. One should just do it right on the install the first time. ;)

> This also allows multi-distro, each with their own encryption password at boot:
> 
> ```
> rpool/arch/...
> rpool/artix/...
> rpool/debian/...

It allows for that yes. It also more-or-less requires it. While the passphrases can be manually set to be the same, they're not the same encryption root. If that is what someone wants, it's a feature. If that is not what someone wants, it's a "bug" (not really, but the user won't love it). So this cuts both ways.

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
ne9z: Since these properties are specific for OS, it might be more appropriate to set these properties on the OS dataset only. So I moved them to the child dataset `rpool/sys`

> Setting `-O acltype=posixacl` enables POSIX ACLs globally. If you do not want this, remove that option

ne9z:
> While the passphrases can be manually set to be the same, they're not the same encryption root.

If one wants to enable encryption for all distros, with the same passphrase, then yes, she has to manually set the same passphrase, this is supported.

Disabling encryption or setting different passphrases is also supported, thus more flexible than shared encroot.

Of course, not sharing the same encroot also means that the master key is not shared, so more secure than shared encroot.

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
ne9z: Moved back.

ne9z:
Scrolling [LUKS faq](https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions#2-setup):
>  On the other hand, an initrd is about as vulnerable to a
competent attacker as a non-encrypted root, so there really is no
security advantage to doing it that way.  An attacker that wants to
compromise your system will just compromise the initrd or the kernel
itself.

So unencrypted boot pool is a security risk, especially on a system with encrypted root pool. The attacker could just modify the initrd unimpeded to intercept the root pool passphrase.

I'm experimenting LUKS 1 encrypted boot pool now, some points:
- GRUB will decrypt LUKS container at boot, passphrase is entered here
- Boot pool MUST be mirror, GRUB can only decrypt one LUKS container at boot
- To avoid entering the passphrase twice, a keyfile is created for LUKS container
- To avoid entering the passphrase twice, root pool should be encrypted with keyfile
- To simplify installation process, default keyfile will be created for root pool, effectively bring back yesterday's text
- Two keys are embedded in initrd
- Remote unlock no longer possible; GRUB does not support this
- Resume hibernation from separate, encrypted swap partition should be possible with [encrypt hook](https://wiki.archlinux.org/index.php/Sd-encrypt#Using_encrypt_hook)

ne9z:
@rlaager: LUKS 1 encrypted boot pool has been added for Arch Linux to mitigate malicious modification of boot pool content.

ne9z:
My fault. Turns out that GRUB can indeed unlock multiple LUKS1 containers at boot.

GRUB unlocks multiple LUKS1
GRUB LUKS1 ZFS boot pool

This works really well with mirrored, LUKS 1 encrypted boot pool.

Removing one disk from mirrored pool also works. GRUB will report one cryptodisk missing and continue to boot.

Currently password must be interactively entered at boot. Some discussion about automating password input for cloud computing is available [here](https://blog.hansenpartnership.com/deploying-encrypted-images-for-confidential-computing/).

ne9z:
So generally LUKS mirrored bpool works with GRUB. If the GRUB prefix disk failed, intervention is required to point GRUB prefix to one of the remaining disks.

ne9z:
I'm somewhat not sure what is the use case here.

If we can ensure the computer can not be physically approached by an attacker, or guarantee the computer would not be used again after such an approach, then unencrypted boot pool is perfectly fine.

Immobile computers, such as server, can be secured away in some cabinet. But for servers separately encrypted datasets are generally preferred over encrypted root. Administration of the latter is more difficult.

Laptops and desktops can be within the reach of attackers. With this scenario, a chain-of-trust should be established. From trusted UEFI firmware (eg. HP Sure Start), GRUB with Secure Boot to encrypted boot and root.

Trusted Cloud Computing has similar requirements to laptops, as discussed [here](https://blog.hansenpartnership.com/deploying-encrypted-images-for-confidential-computing/).

rlaager:
> initrd can be trivially modified to intercept and transfer root pool passphrase.

Two things to remember about security are: 1) What is your threat model? 2) There are always trade-offs.

I'm personally less worried about a motivated attacker trojaning my initrd to capture my passphrase. I'm primarily trying to protect the data at rest. That is, if someone steals my laptop, the data is protected. The thief is far more likely to be stealing my laptop out of a desire for the laptop, not the data. But keeping the data encrypted at rest means I don't need to worry so much about the data being compromised.

There are other environments with other threat models. If I've got Top Secret national security data on my laptop, the threat model is likely very different. But I'm not trying to target that high of a threat model / security environment (nor am I likely competent to do so by myself).

It is certainly possible to do the unlocking with GRUB, using a LUKS-encrypted boot partition of some kind. Then, inside that LUKS-encrypted boot partition, you can store a keyfile which is used to unlock the root filesystem/pool. I even implemented that once in the Ubuntu HOWTO. But I eventually dropped that because a key design goal of mine is to follow the distro (Ubuntu in this case) as closely as possible. I'm trying to make ZFS an option, not second-guess all the distro design decisions. Ubuntu uses an unencrypted initrd and has the initrd (not GRUB) prompt for the passphrase, so that's what I did. Following the distro is less jarring for users; they are presumably happy with their distro choice and just want ZFS. It also makes it far easier and thus far more likely that ZFS support can be integrated into the distro and/or its installer.

I encourage you to follow whatever Arch is doing for non-ZFS full-disk encryption setups. If Arch is doing GRUB unlocking, do that. If Arch is doing initrd unlocking, do that. If you dislike their design decisions, then try to address those in the distro rather than carrying large deltas in the ZFS HOWTO.

ne9z:
@rlaager

Arch doesn't recommend any particular set up - the distro is all about choices. This can be seen at the official [installlation guide](https://wiki.archlinux.org/index.php/Installation_guide#Installation). It lets you choose everything: networking backend, bootloader, and system encryption.

They have docs for [encrypting boot with LUKS](https://wiki.archlinux.org/index.php/Dm-crypt/Encrypting_an_entire_system#Encrypted_boot_partition_(GRUB)). So there's no "delta" from distro design.

Extracting the key in live environment is already documented, and GRUB supports failover (somewhat). Only drawback is missing remote unlock support. But LUKS bpool + keyfile rpool can be converted to unencrypted bpool + password rpool anytime to restore this support.

Gains:
- Protects boot pool from tempering
- Supports hibernation
- Does not require the user to set a strong password at the very beginning
- LUKS master key can be changed online with `cryptsetup-reencrypt`
- No need to destroy rpool if the password is compromised
- Flexible: Can be downgraded to unencrypted bpool + password rpool anytime, online

On a side note, Debian Calamares installer, used in GUI live images, defaults to LUKS encrypted boot partition when  btrfs and encryption is selected.

ne9z:
Reverted. Suggested enablin Secure Boot with signed initramfs to mitigate attack.

ne9z:
@rlaager: I added LUKS bpool as a configurable option after installation for Arch root on ZFS. What do you think?

ne9z:
Well I bricked by EliteBook 820 G3 laptop by using generic tool `KeyTool.efi`. Warning messages added.

ne9z:
> I encourage you to follow whatever Arch is doing for non-ZFS full-disk encryption setups.

Arch Linux is a DIY distro, there's no installer, only a minimal [installation guide](https://wiki.archlinux.org/index.php/installation_guide) exists as framework. Users are also encouraged to choose their own encryption, as evidenced by [docs](https://wiki.archlinux.org/index.php/Dm-crypt/Encrypting_an_entire_system).

In the same spirit of ArchWiki, optional LUKS bpool is documented in guide.

ne9z:
@rlaager: I'm done with the changes. Please review.

PR #138, opened 2021-03-13
updates for 2.0.3

HankB:
remove the redundant symlink operation and fix the path to zpool in the boot mount service file
@rlaager

### DIFF
--- a/docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
@@ -711,7 +711,7 @@ Step 4: System Configuration
          [Service]
          Type=oneshot
          RemainAfterExit=yes
-         ExecStart=/sbin/zpool import -N -o cachefile=none bpool
+         ExecStart=/usr/sbin/zpool import -N -o cachefile=none bpool
          # Work-around to preserve zpool cache:
          ExecStartPre=-/bin/mv /etc/zfs/zpool.cache /etc/zfs/preboot_zpool.cache
          ExecStartPost=-/bin/mv /etc/zfs/preboot_zpool.cache /etc/zfs/zpool.cache
@@ -815,7 +815,6 @@ Step 5: GRUB Installation
      mkdir /etc/zfs/zfs-list.cache
      touch /etc/zfs/zfs-list.cache/bpool
      touch /etc/zfs/zfs-list.cache/rpool
-     ln -s /usr/lib/zfs-linux/zed.d/history_event-zfs-list-cacher.sh /etc/zfs/zed.d
      zed -F &
 
    Verify that ``zed`` updated the cache by making sure these are not empty::


rlaager:
Why is that symlink redundant? Presumably that file does not exist by default, or that ln would fail, since it has no -f.

On the one hand, `/sbin/zpool` is technically correct as written: https://packages.debian.org/buster-backports/amd64/zfsutils-linux/filelist Though I do acknowledge that the HOWTO only creates usrmerge systems, so it's not really problematic to change it either. Is the symlink creating any practical problem?

ne9z:
From [2.0](https://github.com/openzfs/zfs/releases/tag/zfs-2.0.0) onward, zfs-mount-generator is enabled by default on Linux, `/etc/zfs/zed.d/history_event-zfs-list-cacher.sh` exists by default.

rlaager:
Okay, so this is something I need to review (hopefully soon) as part of 2.0 being in buster-backports.

ne9z:
Actually `zfs-list.cache` can be generated without interactively invoking zed, see [here](https://openzfs.github.io/openzfs-docs/Getting%20Started/Arch%20Linux/Arch%20Linux%20Root%20on%20ZFS.html#system-configuration). It's just a few commands.

HankB:
I manually executed the instructions for Debian Buster on ZFS root on real
hardware and found:
1) I did not experience the problem with the path to zpool.
2) zed populated the cache files as desired.
It is not clear to me why I experienced these testing in a VM and I will
continue testing that as I continue to develop my script. One difference is
that I followed the instructions for installing using MBR boot since this
server is too old to support UEFI.

In Step 4 section 8 there is the instruction "Select (using the space bar)
all of the disks (not partitions) in your pool." The menu is not displayed
at this point. I saw later that the command "dpkg-reconfigure grub-pc" will
display that menu. I believe this was covered by later instructions (in the
chroot) and the system did boot normally.

About the issue with the symlink. When I got to that point in the
instructions (Step 5 section 7) This is what I found:
***@***.***:/# ls -l
/usr/lib/zfs-linux/zed.d/history_event-zfs-list-cacher.sh
/etc/zfs/zed.d/history_event-zfs-list-cacher.sh
lrwxrwxrwx 1 root root   57 Mar 13 09:02
/etc/zfs/zed.d/history_event-zfs-list-cacher.sh ->
/usr/lib/zfs-linux/zed.d/history_event-zfs-list-cacher.sh
-rwxr-xr-x 1 root root 2483 Feb 23 08:57
/usr/lib/zfs-linux/zed.d/history_event-zfs-list-cacher.sh
***@***.***:/#

An unrelated comment but regarding ZFS 2.0.x is compression. The
instructions specify lz4. Would it be feasible (and desirable?) to replace
this with an environment variable that could specify lz4 or zstd? (I'm
happy with lz4 since my other systems use that but if I were starting with
ZFS I'd use zstd.)

rlaager:
I'll take your word that the symlink exists by default. I merged that part in 95160f34b69adcbe6ea300f23daa2b36b0adcdb1. The `/sbin/zpool` path is technically correct, and it shouldn't matter. I'm going to be focusing my efforts on getting a Bullseye version of the HOWTO shipped.

Sorry this took so long!

ne9z:
@rlaager It's in [this commit](https://github.com/openzfs/zfs/commit/df834a7ccc6a4bb690f6992273a7345cc56afd6d#diff-ba92c3c3c722f9601d0ab4518afe0531eb0bd96e08844463dda9c0b8143322b1).

PR #141, opened 2021-03-14
Remove duplicate creation of /mnt/run

colmbuckley:
Mounting `/mnt/run` and making `/mnt/run/lock` was already done above; no need to repeat it here.
@rlaager am I missing something?


rlaager:
Thanks for letting me know! I messed this up (for Debian, but not Ubuntu) in the various flailing about with /run/lock issues.

PR #144, opened 2021-03-15
Fix command to unmount ZFS filesystems

mbevand:
Hi @rlaager 
I just followed the guide to install debian buster on ZFS root and noticed an error in the instructions that's present in all the instructions for Debian, Ubuntu, and openSUSE: the command to unmount ZFS filesystems was attempting to unmount all filesystems except the ZFS ones (grep -v). I believe the opposite should be done (grep without -v).

Signed-off-by: Marc Bevand <m.bevand@gmail.com>

### DIFF
--- a/docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Buster Root on ZFS.rst
@@ -866,7 +866,7 @@ Step 6: First Boot
 #. Run these commands in the LiveCD environment to unmount all
    filesystems::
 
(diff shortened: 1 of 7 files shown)


rlaager:
I think that's correct as written. ZFS filesystems will be unmounted when the pool is exported. We need to unmount all the other filesystems, like the bind mounted ones, the ESP, /mnt/run/lock, etc. I suppose there's a legitimate question of whether the grep is necessary at all (i.e. can we just unmount the ZFS ones at the same time), but I'm pretty confident we do not want to reverse the current logic.

mbevand:
I think you were right. Sorry for the false report !

PR #146, opened 2021-03-17
Arch Linux: deprecate cachefile; import pools by scan

ne9z:
The guide currently uses cachefile for import, which is [planned for deprecation](https://github.com/zfsonlinux/zfs/issues/1035) and can not be updated without rebuilding initrd. This pull request adapts the guide to **import by scan**.

Other changes are:
Arch Linux: include var/log,var/spool,var/tmp in system dataset, this is recommended by Arch
Arch Linux: install efibootmgr by default

### DIFF
--- a/docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
+++ b/docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
@@ -291,8 +291,8 @@ System Installation
         # mirror \
         ...
 
-#. When creating pools, for multi-disk installation, you can also use other topologies
-   such as ``raidz1``, ``raidz2`` and ``raidz3``.
+#. When creating pools, for multi-disk installation, other topologies
+   such as ``raidz1``, ``raidz2`` and ``raidz3`` are also available.
 
 #. Create boot pool::
 
@@ -468,9 +468,8 @@ System Installation
      encrypt boot pool and enable Secure Boot, as described in
      the optional configuration section. This will prevent attacks to
      initrd.
-     However, GRUB as of 2.04 requires interactively entering password,
-     you must phsically type in the passwords at boot time,
-     or else the computer will not boot.
+     However, GRUB as of 2.04 requires password to be interactively
+     typed in at boot time, or else the computer will not boot.
 
      Create dataset::
 
@@ -508,13 +507,12 @@ System Installation
         zfs create -o canmount=off rpool_$INST_UUID/sys/DATA/default/$i
     done
 
-    for i in {home,root,srv,usr/local,var/log,var/spool,var/tmp};
+    for i in {home,root,srv,usr/local,var/log,var/spool};
     do
         zfs create -o canmount=on rpool_$INST_UUID/sys/DATA/default/$i
     done
 
     chmod 750 /mnt/root
-    chmod 1777 /mnt/var/tmp
 
 #. Create optional user data datasets to omit data from rollback::
 
@@ -544,7 +542,7 @@ System Installation
 
 #. Install base packages::
 
-     pacstrap /mnt base vi mandoc grub
+     pacstrap /mnt base vi mandoc grub efibootmgr
 
 #. Check compatible kernel version::
 
@@ -571,15 +569,18 @@ System Installation
 
      pacstrap /mnt linux-firmware intel-ucode amd-ucode
 
-#. If you boot your computer with EFI::
-
-     pacstrap /mnt efibootmgr
-
 #. For other optional packages,
    see `ArchWiki <https://wiki.archlinux.org/index.php/Installation_guide#Installation>`__.
 
 System Configuration
 --------------------
+
+#. Set `mkinitcpio zfs hook scan path
+   <https://github.com/archzfs/archzfs/blob/master/src/zfs-utils/zfs-utils.initcpio.install>`__::
+
+    echo 'GRUB_CMDLINE_LINUX="zfs_import_dir=/dev/disk/by-id"' >> /mnt/etc/default/grub
+
+   If using ``/dev/disk/by-path``, set it here.
 
 #. Generate list of datasets for ``zfs-mount-generator`` to mount them at boot::
 
@@ -611,7 +612,7 @@ System Configuration
     echo UUID=$(blkid -s UUID -o value ${DISK[0]}-part1) /boot/efi vfat \
     x-systemd.idle-timeout=1min,x-systemd.automount,noauto,umask=0022,fmask=0022,dmask=0022 0 1 >> /mnt/etc/fstab
 
-   If a swap partition has been created::
+   Add swap. Skip if swap was not created::
 
     for i in ${DISK[@]}; do
      echo swap-${i##*/} ${i}-part4 /dev/urandom swap,cipher=aes-cbc-essiv:sha256,size=256,discard >> /mnt/etc/crypttab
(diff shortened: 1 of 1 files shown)


ne9z:
@rlaager

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
```diff
@@ -508,13 +508,12 @@ System Installation
         zfs create -o canmount=off rpool_$INST_UUID/sys/DATA/default/$i
     done
 
-    for i in {home,root,srv,usr/local,var/log,var/spool,var/tmp};
```
rlaager, line 511: I don't think this is a good idea. /var/log and /var/spool should be separate from the system dataset.

If I roll back my OS dataset, I should not lose user data. /var/spool contains user data like email (if traditional /var/spool/mail is used) and cron jobs. I realize that /var/spool may not be used much on modern, single-user systems, but when it is used, it definitely contains user data.

/var/log is definitely used and definitely contains user data. If I rollback my OS, I do not want to roll back logs. I might need those logs to understand what broke on the system, leading to the rollback. I might also need those logs on a server for security/regulatory reasons.

/var/tmp can go either way. It's not used all that much, so it doesn't matter too much either way. An argument for separating it out is that you can skip snapshotting it. It looks like the current Ubuntu HOWTO no longer separates out /var/tmp. I feel like I used to.

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
```diff
@@ -874,52 +870,46 @@ Optional Configuration
...
-       After=zfs-import.target
+       After=cryptsetup.target
        Before=boot.mount
-       ConditionFileNotEmpty=/etc/zfs/zpool.cache
        ConditionPathIsDirectory=/sys/module/zfs
 
        [Service]
        Type=oneshot
        RemainAfterExit=yes
-       ExecStart=/usr/bin/zpool import -c /etc/zfs/zpool.cache -aN
```
rlaager, line 912: Personally, I don't love the cache file. But it does still seem to be the preferred approach, regardless of how often people have suggested deprecating it. If you're doing this **only** because you saw someone say it was deprecated, you should probably not. If you're doing this intentionally and knowingly, I won't stand in your way. However, make sure you have well tested having a third pool (i.e. something other than bpool or rpool).

rlaager (changes requested):

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
ne9z: With boot environments, OS dataset is never rolled back. If going back to a previous state is needed, just reboot and select the clone. 

If the user needs to investigate the reason why the system is broken, she should mount the dataset then chroot into it. In this case, `/var/log` should preserve the log up to the broken point and should not be written by other boot environments.. If `/var/log` is shared, after chroot, a bunch of newly written messages will pop up in logs, making debugging a little more confusing.

The point is, system state should be preserved & sealed in a clone whenever possible, instead of mixing system states. 

I agree with you regarding `/var/spool`, added back.

rlaager: I agree with the statement about system states. I do not consider /var/log to be system state, nor did the Ubuntu guys at Canonical in their extensive conversations on what is and is not user data. (And they had a bunch of pre-existing experience from making similar distinctions on the now-defunct Ubuntu phone platform.) I very much think that /var/log should be treated as user data.

Pretend I used your approach on my email server. I apply updates. A while later, I notice that something is wrong. I troubleshoot, but ultimately decide I need to go back, at least temporarily, to get things working for customers again. I boot back into the previous boot environment clone (which I'm still calling a "roll back" even though it's not a literal `zfs rollback`). Now everything is working. Then a customer asks me about a message that did not come through. (This is not at all unusual under normal circumstances, but given that _we just had problems_ it's quite likely to occur.) Now I have a big hole in my logs? That's not okay. And sure, I can manually mount the clone, but that's extra work for no good reason.

For another example, substitute a DHCP server. I need a record of all leases for some period of time, so I can deal with abuse complaints. Your approach would again require me to mount the old BE to get access to the old /var/log. That also means I need to keep the old BE for however long my log retention policy requires. But I might not normally want to keep old BEs that long. This becomes a mess very quickly, for practical scenarios that I deal with routinely.

/var/log is user data and needs to be treated as such.

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
ne9z: I deprecated it because zpool.cache causes major headaches when embedded in initrd.

FYI, Arch Linux uses mkinitcpio to generate initrd, and zfs support is added by [a hook](https://github.com/archzfs/archzfs/blob/master/src/zfs-utils/zfs-utils.initcpio.hook). Its behavior is, embed zpool.cache in initramfs if file exists, and import **only** from the embedded zpool.cache file at boot, skipping scanning entirely. If the vdev path or configuration has been changed, and initrd has not been regenerated, initrd will still import the pool in the old vdev configuration on next reboot, which you know, not good.

Example: recently I sold by laptop, [bricked while testing Secure Boot](https://github.com/openzfs/openzfs-docs/pull/127#issuecomment-786976799). I took the NVMe drive out and put it in an external USB3 enclosure to boot it from another computer. This enclosure contains a RealTek controller, which converts the NVMe disk into an ATA one. In `/dev/disk/by-id`, this changed the device name from `nvme-SAMSUNG-...` to `ata-nvme-SAMSUNG-...`. Of course, thanks to the stale cache, rpool failed to import and the system ends up in kernel panic. One must boot up live CD and chroot into the broken system to delete the stale cache, regenerate initrd and force import-by-scanning on next reboot.

rlaager: Sounds like you're doing so intentionally then, so that's fine. Did you test having a third (non-bpool, non-rpool) pool?

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
ne9z: I agree. Added back.

docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.rst
ne9z: Yes, just tested on my server. The third pool imports fine with import-by-scanning.

ne9z: The pool are imported with `-N`, thus not mounted by initrd. The mounting is handled by zfs-mount.service for third pool, zfs-mount-generator for rpool, fstab for bpool.

ne9z: Boot dataset is a tricky one because while it must be must be mounted at boot, it also must be `canmount=noauto` to support boot environments. zfs-mount-generator will not mount `canmount=noauto` at boot, so `mountpoint=legacy` and fstab are used instead.

rlaager:
Seems good to go, but you might as well squash the /var/spool and /var/log changes back into the original commit. There's not much value in moving them out and them moving them right back in. :)

ne9z:
Commits squashed.

PR #149, opened 2021-03-25
Update dRAID Howto.rst

pashford:
Add default value for Parity.


pashford:
This change is based on the example provided.  The example provides no Parity information and is described as creating a single-parity array.  If that example is correct, then the default is single parity.

PR #153, opened 2021-04-16
Add Code of Conduct

jumbi77:
Port over Code-of-Conduct-Page from OpenZFS Wiki (https://openzfs.org/wiki/Code_of_Conduct)

### DIFF
--- a/docs/Code_of_Conduct.rst
+++ b/docs/Code_of_Conduct.rst
@@ -1,0 +1,156 @@
+Code of Conduct
+===============
+
+Overview
+--------
+
+The OpenZFS community values respectful, welcoming behavior towards everyone.
+This enables our members to thrive and contribute, and encourages new
+participants to join our community.
+
+Although most experiences within the community are positive, we all have moments
+when we are unaware of our tone or are too busy to choose our words carefully.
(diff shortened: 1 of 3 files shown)


gmelikov (changes requested):
Thank you for a contribution!

CoC is not a technical document, our intention with documentation is that this repo has mainly technical parts, while openzfs.org wiki has community-related things. I propose to stay with this way.

gmelikov:
So, we don't plan to move all openzfs.org content, and I'll close this one. If we change this difference - we can get back to this PR.

jumbi77:
Thanks for the quick reply - okay, no problem. However, one downside is that the wiki is not editable for the public and that it is maintained irregularly. IMHO to make openzfs-docs (sphinx doc) the central place for information and documentation about OpenZFS is worth considering (easy editable, more modern looking, no hosting overhead, etc...). It would also be much more uniform for the enduser instead of currently having zfsonlinux.org, community wiki (open-zfs.org), technical sphinx doc etc.

gmelikov:
@jumbi77 the future plan is at least to place openzfs-docs somewhere on openzfs.org domain too, as a subdomain or subdirectory, so community content and technical docs will be on one resource.

PR #159, opened 2021-05-01
Update instructions for Ubuntu on Raspberry Pi for 20.04.2 release.

fernandocarolo:
Update the getting started instructions for Ubuntu on a Raspberry Pi to use the 20.04.2 LTS release and add `-a` flag to the `wipefs` command, to make sure it will erase any filesystem header from previous use of the SD card.
@rlaager


rlaager:
Pedantically: these should be two commits, since they are logically unrelated.

But I'm just going to merge it as is.

Thanks!

PR #164, opened 2021-05-12
Arch Linux: split guide to separate pages

ne9z:
The root on ZFS guide has become quite long and thus difficult to navigate.

This commit split the guide into separate pages.
Navigation is achieved with built-in `prev` & `next` buttons.

@rlaager 


gmelikov:
IMHO one page is better because you shouldn't change pages every time and can just scroll. Navigation works via same Table of Contents and in addition you can search through whole page.

ne9z:
I think this is a matter of personal taste.

- I myself find it to be quite tiresome to scroll on a long page, either with touch pad or trackball. This is the reason why I split sections into pages.
- With separate pages, user can also skip the "optional configuration" section with a single click, instead of scrolling down tons of irrelevant text. FYI, this section occupies ~20% of the line count.
- Page would appear to be less bloated when I add new content :-)

I've replaced the content of current page with an index, so the link should still work.

Separate pages also appears to be easier to navigate, thanks to the sidebar. With a single page, one needs to return to the top of the page to get table of contents. An image demo is attached below:
20210512_20h25m32s_grim

gmelikov:
That was just my IMHO, so let's try it on Arch:)

ne9z:
Maybe we can merge this first and wait for user feedback?

ne9z:
@rlaager Sorry for being pushy, but I really would like to get this PR reviewed and hopefully merged.

There are also several fixes, in addition to the changes given above.

rlaager:
I didn't really look at this, but it's merged now.

PR #166, opened 2021-05-22
Updated and corrected build instructions.

rincebrain:
While adding python{,3}-packaging to the build requirements, I
suspected that the build instructions hadn't actually been tested
recently. So I checked, found problems, and corrected them.

Briefly:
* --skip-broken is needed in a number of commands to ignore the absence of EPEL/powertools on Fedora
* Broke out the install commands that specifically required EPEL/powertools; the alternative was breaking out the command to install EPEL, I have no strong opinion on which one to prefer.
* I had to break out the Debian/Ubuntu install step because Debian calls it linux-perf-KERNELVER with a linux-perf metapackage, while Ubuntu calls it linux-tools-common, neither has a compat package for the other's name, and apt appears to have no equivalent for --skip-broken to just specify both.
(The alternative of putting a shared Debian/Ubuntu command and then additional Debian/Ubuntu specific steps below it seemed to me to be likely to cause confusion; I can do that instead if people disagree.)
* EPEL is needed for python-packaging on EL7 (powertools, which ships disabled with the OS, is needed on CentOS8; I do not have an actual RHEL8 system to examine the state there)
* We still need --skip-broken for the command installing pax because it's (AFAICT) entirely absent on EL8

I've tested these instructions with:
* Ubuntu 21.04 x64
* Fedora 34 x64
* Debian 11 mips64el
* CentOS 8 x64
* CentOS 7 x64

And it successfully built {deb,rpm}-{kmod,dkms,utils} as appropriate (except on mips64el, but that seems to just be broken on the platform, not a problem of a missing dependency)

(If openzfs/zfs#12102 lands, I can instead drop the -packaging additions and rewrite the checks from openzfs/zfs#12073 that are the whole reason for adding the requirement...)

### DIFF
--- a/docs/Developer Resources/Building ZFS.rst
+++ b/docs/Developer Resources/Building ZFS.rst
@@ -16,7 +16,7 @@ There are two main components in this repository:
    vast majority of the core OpenZFS code is self-contained and can be
    used without modification.
 
-- **SPL**: The SPL is thin shim layer which is responsible for
+- **SPL**: The SPL is a thin shim layer which is responsible for
    implementing the fundamental interfaces required by OpenZFS. It's
    this layer which allows OpenZFS to be used across multiple
    platforms. SPL used to be maintained in a separate repository, but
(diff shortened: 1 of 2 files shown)


gmelikov (approved):
LGTM, thanks!

PR #177, opened 2021-06-18
Add Fedora Root on ZFS support

ne9z:
This PR adds Fedora Root on ZFS support, with GRUB as bootloader.

For traditional installation, a fallback to LTS kernel has been added.

@rlaager @cheesycod


docs/Getting Started/index.rst
```diff
@@ -13,7 +13,7 @@ documentation <https://pthree.org/2012/04/17/install-zfs-on-debian-gnulinux/>`__
 
    Arch Linux/index
    Debian/index
-   Fedora
```
gmelikov, line 16: Let's do the same as with https://github.com/openzfs/openzfs-docs/pull/180/files#r681526008

ne9z: Thanks for the notice, fixed.

gmelikov:
@ne9z as we don't have active contributors with Fedora as main distro, I'll merge it as is. Thank you!

techhazard:
@ne9z I previously installed Fedora 34 on a ZFS root by using the Ubunto 20.04 guide and adapting where necessary. And I saw that you added a guide specific to Fedora, super!

But the different dataset layout is something that is not clear to me.
Based upon the Ubuntu guide, my layout for `rpool` is the following:

```
NAME
rpool
rpool/ROOT
rpool/ROOT/fedora_34
rpool/ROOT/fedora_34/etc
rpool/ROOT/fedora_34/srv
rpool/ROOT/fedora_34/usr
rpool/ROOT/fedora_34/usr/local
rpool/ROOT/fedora_34/var
rpool/ROOT/fedora_34/var/lib
rpool/ROOT/fedora_34/var/lib/...
rpool/ROOT/fedora_34/var/log
rpool/ROOT/fedora_34/var/spool
rpool/ROOT/fedora_34/var/www
rpool/USERDATA
rpool/USERDATA/root
rpool/USERDATA/techhazard
```

This way I can add a `rpool/ROOT/fedora_35` installation in the future.
Also, my `rpool` is currently my encryptionroot
However, based on the suggested dataset layout it seems it would be something like this:

```
NAME
rpool
rpool/sys
rpool/sys/ROOT
rpool/sys/ROOT/fedora_34
rpool/sys/ROOT/fedora_34/etc
rpool/sys/ROOT/fedora_34/srv
rpool/sys/ROOT/fedora_34/usr
rpool/sys/ROOT/fedora_34/usr/local
rpool/sys/ROOT/fedora_34/var
rpool/sys/ROOT/fedora_34/var/lib
rpool/sys/ROOT/fedora_34/var/lib/...
rpool/sys/ROOT/fedora_34/var/log
rpool/sys/ROOT/fedora_34/var/spool
rpool/sys/ROOT/fedora_34/var/www
rpool/sys/DATA
rpool/sys/DATA/default
rpool/sys/DATA/default/home
rpool/sys/DATA/default/home/root
rpool/sys/DATA/default/home/techhazard
```

But looking at the actual installation steps, it looks like I will end up with this:
```
NAME
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/ROOT
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/ROOT/default
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/etc
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/srv
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/usr
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/usr/local
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/var
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/var/lib
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/var/lib/...
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/var/log
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/var/spool
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/var/www
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/home
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/home/root
rpool_a75c1d37-69fd-4b0d-b7ce-d8df85afc730/fedora/DATA/default/home/techhazard
```

For me this isn't as clear as the more simple setup from the ubuntu guide, and raises several questions:
- why the added UUID to the pool name? This makes it way harder to import the pool when you need to do manual imports when the bootloader/initramfs fails.
- why the difference between the [Dataset Layout as described in the Overview section](https://openzfs.github.io/openzfs-docs/Getting%20Started/Fedora/Root%20on%20ZFS/0-overview.html#dataset-layout) and the `rpool_$INST_UUID/$INST_ID/...` layout as used in the rest of the guide?
- why is `rpool/sys` suggested as an encryptionroot and not `rpool`?
- more generally: why the `rpool/sys` subdataset? It doesn't seem to add much.
- I'm quite perplexed why you would want to have `/` on `rpool/fedora/ROOT/default` but then have all the subdirectories on `rpool/fedora/DATA/default/...`. I'm sure there's a reason for it, I just can't see it.
- What are the benefits of using the approach from the new Fedora guide as opposed to the (IMO simpler) layout of the Ubuntu guide?
  - Is it easier to make certain snapshots this way?
  - Is upgrading to a new version of fedora easier?
  - Does this work better with tools like [zfsbootmenu](https://github.com/zbm-dev/zfsbootmenu)?

I'd love to see some reasoning added to the guide to explain the choices made for this pool layout.


Also, secure boot and ZFS is possible, because I'm using it; but it requires a lot more work (add your own Secure Boot keys and sign the kernel module and efi binaries)

ne9z:
> But the different dataset layout is something that is not clear to me.

This guide is based on the Arch Linux guide.

> This way I can add a `rpool/ROOT/fedora_35` installation in the future.

The `/sys/` container is specifically designated for this purpose, that is, your system would be
```
rpool/fed35
rpool/fed36
```
The dataset you are currently using is reserved for boot environments:
```
rpool/fed35/ROOT/be0
rpool/fed35/ROOT/be1
```
> Also, my `rpool` is currently my encryptionroot

This is inflexible, see [here](https://github.com/openzfs/openzfs-docs/pull/127#issuecomment-785049655).

> But looking at the actual installation steps, it looks like I will end up with this:

The UUID is a 6-char alphanumeric string, like `rpool_abc123`.

>     * why the added UUID to the pool name? This makes it way harder to import the pool when you need to do manual imports when the bootloader/initramfs fails.

First, I do testing with VMs. There are several Root on ZFS installations on multiple disks, all of them attached to the same VM. If I do not use the UUIDs, there will be multiple `rpool`, and none of them will be imported at boot due to the duplicating name.

With boot environments, we aims to eliminate the need to do manual imports altogether. You can try it in a VM.

>     * why the difference between the [Dataset Layout as described in the Overview section](https://openzfs.github.io/openzfs-docs/Getting%20Started/Fedora/Root%20on%20ZFS/0-overview.html#dataset-layout) and the `rpool_$INST_UUID/$INST_ID/...` layout as used in the rest of the guide?

Because the table, in reStructuredText, is quite difficult to edit. To save some effort, and to avoid cluttering the table, variables are omitted in the introduction.

>     * why is `rpool/sys` suggested as an encryptionroot and not `rpool`?
>     * more generally: why the `rpool/sys` subdataset? It doesn't seem to add much.

See above.

>     * I'm quite perplexed why you would want to have `/` on `rpool/fedora/ROOT/default` but then have all the subdirectories on `rpool/fedora/DATA/default/...`. I'm sure there's a reason for it, I just can't see it.

This avoids cluttering up the BE container.

```
ROOT/be0
ROOT/be1
DATA/var/lib
``` 
not
```
ROOT/be0
ROOT/be0/var/lib
ROOT/be1
```
Also see #127.

>     * What are the benefits of using the approach from the new Fedora guide as opposed to the (IMO simpler) layout of the Ubuntu guide?
>       
>       * Is it easier to make certain snapshots this way?
>       * Is upgrading to a new version of fedora easier?

See above. This does not affect upgrade in anyway.

>       * Does this work better with tools like [zfsbootmenu](https://github.com/zbm-dev/zfsbootmenu)?

This guide is compatible with zbm, but not specifically designed for it. It is designed for [bieaz](https://gitlab.com/m_zhou/bieaz).

> Also, secure boot and ZFS is possible, because I'm using it; but it requires a lot more work (add your own Secure Boot keys and sign the kernel module and efi binaries)

I'm aware of that, see [here](https://github.com/openzfs/openzfs-docs/pull/127#issuecomment-786976799). It's just too complex to be included in this guide.

To sum up, I suggest you to give this guide a try in a VM.

techhazard:
Thanks for the reply!

> > Also, my rpool is currently my encryptionroot
>
> This is inflexible, see here:
>> Actually I think enabling encryption on the pool dataset rpool is quite inflexible. The only way to get rid of the encryption or change any readonly dataset property is to destroy and rebuild the pool

Yes, that makes sense, I'll keep that in mind in the future.

> First, I do testing with VMs. There are several Root on ZFS installations on multiple disks, all of them attached to the same VM. If I do not use the UUIDs, there will be multiple rpool, and none of them will be imported at boot due to the duplicating name.

It makes a lot more sense when using multiple root pools, I agree. But for more "regular" solutions I don't think thats necessary, so maybe add a note in the guide about it? (Also, [UUID's are generally 36 characters](https://en.wikipedia.org/wiki/Universally_unique_identifier), as you probably know; so I didn't even see that you only used 6 characters instead.)

> With boot environments, we aims to eliminate the need to do manual imports altogether. You can try it in a VM.

I agree that it _shouldn't_ be necessary to import the pool manually, and I almost never have to. But sometimes the boot process breaks, and typing a whole UUID is a bit much, but if it's just 6 characters than that isn't as bad.

> This guide is compatible with zbm, but not specifically designed for it. It is designed for bieaz.

I only mentioned zbm because I recently ran into it, and still wanted to try it out. I'll take a look at bieaz too. I just saw a lot of choices were made regarding the pool layout, and no obvious answers.

> >  Also, secure boot and ZFS is possible, because I'm using it; but it requires a lot more work (add your own Secure Boot keys and sign the kernel module and efi binaries)
>
> I'm aware of that, see here. It's just too complex to be included in this guide.

I get that it's too complex for this guide, but maybe add a note that it's technically possible (maybe point to the [references I used here](https://gist.github.com/techhazard/e5feca0535317d2a94a6f086433d3a25#references)?) 


sidenote: I removed grub altogether and use direct efi boot (with `efibootmgr`) or systemd-boot (with `bootctl`), depending on the system.  That way I only have an EFI partition and don't need a separate boot pool or separate boot environments/BE (I'm still not quite sure what is meant by that, I don't think it's initramfs?).

To recap, I don't really disagree  with the choices that are made in the new guide, but I think it can be improved a lot by adding some explainations and/or references in the guide. Like the reason why the pool isn't the encryptionroot, and the reason(s) for a separate `/boot` pool and separate boot-environment datasets. And a mention of the ArchWiki guide that it's based on.
The switch from `rpool/sys/...` to `rpool_$INST_UUID/$INST_ID` is not really explained, so maybe  add a note on that as well?

ne9z:
@techhazard 

> so maybe add a note in the guide about it
I would recommend everyone to use the unique identifier. Another thing is the convenience of testing.

I have to test the guide, including every option, every 2 months. Removing the UUID will increase my workload.

> as you probably know; so I didn't even see that you only used 6 characters instead

This one is originated from Ubuntu, see [Ubuntu guide](https://openzfs.github.io/openzfs-docs/Getting%20Started/Ubuntu/Ubuntu%2020.04%20Root%20on%20ZFS.html) and search `UUID` in page. Although they used it for BEs only.

> add a note that it's technically possible

Previously there was a note to use PreLoader. Some people blindly copy-pasted the code and enabled Secure Boot by accident. There are also folks like me who straight up broken their computer with the process. I guess those who need SB will look elsewhere for solutions.

> I'll take a look at bieaz too.
> I removed grub altogether

Unfortunately bieaz depends on a functional GRUB installation, as described in the guide. So you might need a fresh install to start using it.

> don't need a separate boot pool or separate boot environments/BE

It works like so:

You are about to perform an upgrade, before the upgrade, run
```
bieaz create be0
bieaz label be0 'pre-upgrade' # optional
```

A new item will appear in GRUB menu to let you to go back to the state before the upgrade.

The complicated layout is mostly for separating user data from BE. See the guide for detail.

> separate /boot pool and separate boot-environment datasets

The explanations are in the guide already, if you look closely. ;-)

PR #179, opened 2021-06-19
Add RHEL-compatible distro Root on ZFS support

ne9z:
This PR cleans up the existing page for RHEL and adds support of Root on ZFS.

CentOS can no longer represent the situation of RHEL-compatible derivatives, AlmaLinux is used instead.

Actually, supporting RHEL is easier than Fedora due to the stable kernel version.


ne9z:
``grub2-probe`` in GRUB 2.02 is incompatible with zfs. RHEL 8 ships GRUB 2.02 and a newer release will take at least several years.

PR #181, opened 2021-06-23
Add rpool to /etc/zfs/zpool.cache

RemyNoulin:
This change makes the system boot completely without the need to run 'zpool import -f' at the first boot.


gmelikov:
Looks like it's not mandatory if you did export the pool after install just before first reboot. Tested on personal machine, I don't set cachefile for root pool. Did you export pools as instruction says?

RemyNoulin:
I close this pull request, I found that I didn't stop zed ('fg + ctrl-c') in step 5

PR #188, opened 2021-08-10
Fedora: Use compatibility property for bpool; Arch: Update GRUB config for Live image

ne9z:
OpenZFS has recently been updated to version 2.1.0 on Fedora, the need to specify grub2 compatible pool features is therefore removed.

A couple of fixes for booting Arch Linux live image directly from EFI system partition.

@gmelikov 


gmelikov:
@ne9z let's split different changes by commits next time.

gmelikov (approved):

PR #193, opened 2021-08-22
Added Ubuntu 20.04 Root on ZFS on external USB for Raspberry Pi guide

gador:
Added a guide to install Ubuntu 20.04.2 on external USB drives with ZFS for the Raspberry Pi 4. 

@rlaager: I'd appreciate any feedback, thanks!

Signed-off-by: Florian Brandes <florian.brandes@posteo.de>

### DIFF
--- a/docs/Getting Started/Ubuntu/Ubuntu 20.04 Root on ZFS on external USB for Raspberry Pi.rst
+++ b/docs/Getting Started/Ubuntu/Ubuntu 20.04 Root on ZFS on external USB for Raspberry Pi.rst
@@ -1,0 +1,880 @@
+.. highlight:: sh
+
+Ubuntu 20.04 Root on ZFS on external USB for Raspberry Pi
+=========================================================
+
+.. contents:: Table of Contents
+  :local:
+
+Overview
+--------
+
+Caution
+~~~~~~~
+
+- This HOWTO uses an external HDD or SSD connected via USB.
+- Backup your data. Any existing data will be lost.
+
+System Requirements
+~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


gmelikov:
Do we really need to have another Howto for this case? Can we just add notes in main one?

gador:
I understand where you are coming from. Unfortunately, it is quite different from ZFS on raspberry pi sd card installation:

-  in the SD Card tutorial, ZFS is installed on a third partition of the sd card and later expanded, after the ext4 partition of the original image isn't needed anymore. My guide doesn't do that. It installs ZFS on an external USB drive independent of the raspberry pi. There is no need to do anything with the SD card.  Also there is quite a bit to change to make the raspberry pi boot from an external drive.

 It also is majorly different than the Ubuntu 20.04 guide:

- You cannot use the installer for ZFS for the rasperry pi
- There is no "live cd" for the raspberry pi. One could use a normal raspberry pi image and attach the usb drives, but all the other raspberry pi specific things would need to be added.

To summarize: It just is a different approach and In my opinion it would confuse the reader more, when all kind of notes would need to be added to either of the other two guides.

gmelikov:
@rlaager could you please look at it, if you have some time.

rlaager:
I merged a different version of this in f60e80083d96b948baad711f6faa71001db2b12c. If there are more changes necessary, please open a new PR based on current master.

I'm very sorry I didn't get to this. Life got very busy for me and I haven't had the time to work on ZFS (or other FOSS stuff) as much as I'd like. I know it really sucks to contribute a PR and have it languish, and I'm genuinely sorry that I did that to you.

PR #199, opened 2021-09-06
Update: SATA drives with power failure protection

DavidPesticcio:
Old location: http://blog.nordeus.com/dev-ops/power-failure-testing-with-ssds.htm (404 Not Found)

Present location: https://engineering.nordeus.com/power-failure-testing-with-ssds


gmelikov (approved):
Thank you!

PR #200, opened 2021-09-07
Updated compatibility_matrix.py for zpool-features.7

rincebrain:
In OpenZFS 2.1 and newer, zpool-features got refactored and renamed, so it no longer shows up in the feature matrix.

This corrects that.


gmelikov (approved):
Thank you, looks good!

PR #202, opened 2021-09-08
Always use mirrored vdev for boot pool

ne9z:
I just stumbled across [this issue](https://github.com/openzfs/zfs/pull/11763), 
where discussion concluded that GRUB does not support indirect vdevs and will not be able to read the pool even if there's enough redundancy.

This reminded me that we must ensure boot pool is always readable by GRUB and unaffected by any new pool feature what-so-ever. Or else the computer will not be able to boot at all.

This commit changes the guide to use mirrored vdev for boot pool under all circumstances except single disk installation.

@gmelikov

### DIFF
--- a/docs/Getting Started/NixOS/index.rst
+++ b/docs/Getting Started/NixOS/index.rst
@@ -89,4 +89,4 @@ Contribute
     sensible-browser _build/html/index.html
 
 #. ``git commit --signoff`` to a branch, ``git push``, and create a pull
-   request. Mention @rlaager.
+   request. Mention @ne9z.
--- a/docs/Getting Started/NixOS/index.rst
+++ b/docs/Getting Started/NixOS/index.rst
@@ -62,3 +62,31 @@ An installation guide is available.
   :glob:
 
   Root on ZFS/*
+
+Contribute
+----------
+#. Fork and clone `this repo <https://github.com/openzfs/openzfs-docs>`__.
+
+#. Launch an ephemeral nix-shell with the following packages::
+
+    nix-shell -p python39 python39Packages.pip gnumake \
+        python39Packages.setuptools python39Packages.sphinx
+
+#. Create python virtual environment and install packages::
+
+    cd openzfs-docs
+    python -m venv .venv
+    source .venv/bin/activate
+
+    pip install -r docs/requirements.txt
+
+#. Make your changes.
+
+#. Test::
+
+    cd docs
+    make html
+    sensible-browser _build/html/index.html
+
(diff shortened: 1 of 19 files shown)


ne9z:
An Ubuntu user also reported a problem related to indirect vdev [on reddit](https://www.reddit.com/r/zfs/comments/mo599z/after_drive_removal_indirect1_vdev_presumably/). Mirrored boot pool will prevent this from happening.

richardelling:
FWIW, field data from Backblaze shows surprisingly terrible reliability in HDD boot drives.  SSDs fare much better. https://www.backblaze.com/blog/backblaze-drive-stats-for-q2-2021/

ne9z:
Unfortunately that is something outside the scope of the guide. ZFS should improve the reliability of booting the system regardless of HDD or SDD-backed storage.

Personally I'm using a RAID-Z2 setup with 4 HDDs on my 4-bay Dell T20 server.

ne9z:
Edit: I mean, everyone knows performance and reliability is superior compared to HDDs. But not every reader of the guides have the opportunity to upgrade their setups, be it budget constraints, limited number of drive bays, etc.

A note recommending SSD as main storage might be appropriate, but the guide should aim at improving reliability on all platforms.

richardelling:
the point is that HDD boot devices are particularly unreliable and you should mirror them. Even the SSD reliability isn't as good as enterprise class bulk storage.

ne9z:
Yes, this PR will make mirrored vdev default for multi-disk boot pools. Boot pools, default to 4GB in the guide, only holds initrd, which is between 50~100MB in size. Free space is not of particular concern here.

Ideally mirrored vdev should also be used for root pool, as you have pointed out above, but then there's the trade-off between usable space and redundancy. Thus this choice is left to the user to make.

docs/Getting Started/NixOS/index.rst
```diff
@@ -62,3 +62,31 @@ An installation guide is available.
...
+#. Make your changes.
+
+#. Test::
+
+    cd docs
+    make html
+    sensible-browser _build/html/index.html
+
+#. ``git commit --signoff`` to a branch, ``git push``, and create a pull
+   request. Mention @rlaager.
```
gmelikov, line 92: Maybe mention you, @ne9z ?

ne9z: Replaced.

ne9z:
Done. Thanks for the speedy reply!

gmelikov (approved):
LGTM

PR #203, opened 2021-09-08
Add warning about containers

rincebrain:
A couple people have come by IRC having issues where the root cause was that they were stuck on the CDDL/GPL prompt from apt never rendering because their container runtime console wasn't being detected as dumb/noninteractive by apt, but wasn't capable of displaying the prompt.

So let's add a warning about that, shall we?


ne9z:
Nice addition, however with Debian's strong stance on Free Software, shouldn't the prompt default to false/disagree? ;-) Kind of making the prompt an unnecessary annoyance.

gmelikov (approved):
Thank you!

PR #207, opened 2021-09-16
Update Debian apt install command

chrisbanes:
It assumes that the host arch is amd64, which isn't always true.


rlaager:
@chrisbanes Why did you end up closing this? Your change seems sane at first glance.

rlaager:
I suppose amd64 is all that's supported by the instructions. (Obviously, the Raspberry Pi is a different deal.) So this change probably doesn't make a huge difference either way.

Issue #157, opened 2021-04-29
Hibernation


josephnz:
@ne9z, I have the following issue with the Arch Linux Root on ZFS HOWTO:
Can you please include the steps to enable hibernation with encrypted bootpool ?

mskarbek:
https://twitter.com/nabijaczleweli/status/1378679960408371207
@nabijaczleweli

nabijaczleweli:
heh, yeah. i say this with as much sincerity as i can muster: if you have to ask, the answer is "you can't resume off zvols"

(and if you recreate what i did there, you have a fifty-fifty chance of it either not working or raking your pool each time you try if you so much as breathe at it wrong. I permanently damaged that pool more times than I can count, so, I repeat: do *NOT* do it on a pool you can't can)

josephnz:
just to clarify, I have a separate swap partition as described in the document. Are you saying it's still a no go? If so perhaps the document should be modified. For eg: currently it says 
"Adjust the swap partition size to your needs. If **hibernation is needed**, swap size should be same or larger than RAM"

ne9z:
If you have created a separate swap partition during installation and enabled encryption, you need to 

- encrypt boot pool with LUKS1, [see here](https://openzfs.github.io/openzfs-docs/Getting Started/Arch Linux/Arch Linux Root on ZFS.html#optional-configuration)
- encrypt swap partition with LUKS2 and add kernel cmdlines, [see here](https://github.com/openzfs/openzfs-docs/blob/e70da303a3dfee4a279ba9f3c0e51549ff3fdcec/docs/Getting%20Started/Arch%20Linux/Arch%20Linux%20Root%20on%20ZFS.rst#hibernation)

I removed the hibernation section as the guide is getting too long ;-)

ne9z:
@nabijaczleweli I guess you are doing swap on zvol.

- zvol in general is extremely patchy, see [1](https://github.com/openzfs/zfs/issues/11407) and other issues, it should be avoided at all cost.
- swap on zvol is known to crash the system, see [here](https://github.com/openzfs/zfs/issues/7734)
- resume hibernation from a pool will corrupt it, this is [known](https://github.com/openzfs/zfs/issues/260)

[...]

PR #208, opened 2021-09-18
Add `dnf remove -y zfs-fuse` step to fedora preparation doc

infogulch:
I followed the preparation doc for installing zfs on root on fedora for the first time, but got an error at the `dnf install -y zfs` step:

```
# dnf install -y zfs
ZFS on Linux for Fedora 34                                                                                      142 kB/s | 199 kB     00:01
Last metadata expiration check: 0:00:01 ago on Fri 17 Sep 2021 10:29:51 PM EDT.
Error:
 Problem: problem with installed package zfs-fuse-0.7.2.2-18.fc34.x86_64
  - package zfs-2.0.4-1.fc34.x86_64 conflicts with zfs-fuse provided by zfs-fuse-0.7.2.2-18.fc34.x86_64
  - conflicting requests
  - package zfs-2.0.5-1.fc34.x86_64 conflicts with zfs-fuse provided by zfs-fuse-0.7.2.2-18.fc34.x86_64
  - package zfs-2.1.0-1.fc34.x86_64 conflicts with zfs-fuse provided by zfs-fuse-0.7.2.2-18.fc34.x86_64
  - package zfs-2.1.1-1.fc34.x86_64 conflicts with zfs-fuse provided by zfs-fuse-0.7.2.2-18.fc34.x86_64
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages)
```

I was able to work around this by following the first step in [index.rst](https://openzfs.github.io/openzfs-docs/Getting%20Started/Fedora/index.html#testing-repo) to `dnf remove -y zfs-fuse` first.

This change makes the preparation document align with the same recommendation in index.rst. I added this step as late as possible in the list of steps that still prevents the issue I encountered.

### DIFF
--- a/docs/Getting Started/Fedora/Root on ZFS/1-preparation.rst
+++ b/docs/Getting Started/Fedora/Root on ZFS/1-preparation.rst
@@ -38,6 +38,10 @@ Preparation
 
     dnf install -y https://zfsonlinux.org/fedora/zfs-release.fc${VERSION_ID}.noarch.rpm
 
+#. If zfs-fuse from official Fedora repo is installed, remove it first. It is not maintained and should not be used under any circumstance::
+
+    dnf remove -y zfs-fuse
+
 #. Install ZFS packages::
 
     dnf install -y zfs


ne9z:
Thanks for the tip! However, how did `zfs-fuse` got into your live environment in the first place?

As far as I know, this particular package should not be present in a freshly-booted live image.

infogulch:
That's a fair question, but I'm afraid my answer will simply contradict you: I'm certain I just flashed the linked Fedora Workstation (GNOME) iso, [Fedora-Workstation-Live-x86_64-34-1.2.iso](https://mirror.genesisadaptive.com/fedora/linux//releases/34/Workstation/x86_64/iso/), onto a usb drive, booted it, and followed each step verbatim. I got straight to business, no poking around or anything (though I was tempted since I hadn't seen Gnome 40 before).

The checksum of the iso on disk matches the CHECKSUM file and begins with `865c4457`. It doesn't look like the iso has been updated since April (much older than your doc). I'm not sure what gives. Perhaps the default installed packages are different between gnome and spins?

ne9z:
So that's certainly a fault on my part. I'm the author of this and the
RHEL guide, and I have only tested with the more lightweight LXDE spin.

Your message is the first one I received since the guide published.
Have you encountered other difficulties during installation?

By the way, the link to the boot environment manager (also written by
me) has been updated, but the pull request hasn't been merged yet. The
new link is here:
https://nu8.org/pages/projects/bieaz/

Or simply:

```
dnf copr enable -y m0p/bieaz
dnf install -y bieaz python3-dnf-plugin-rozb3
```

This tool will let you rollback system updates. For each DNF
transaction, a pair of pre/post snapshots are created.

infogulch:
> I'm the author of this and the RHEL guide, and I have only tested with the more lightweight LXDE spin.

That's a totally understandable thing to overlook; luckily it was no more than a blip for me since I had seen the other page that recommended removing zfs-fuse and I connected the dots right away. I opened the PR just to knock off that little sharp edge for any that came after me. I did notice that you were the author from the git history of the file. Thanks for writing up this guide, it was very helpful -- targeting a wider audience by including the gnome iso was a good move imo despite this little issue.

> bieaz

bieaz looks pretty slick, I'll have to check that out later. Using snapshots to enable rolling back system updates (or even just listing out the specific changes made with `zfs diff`) is exactly the kind of thing that I'm excited to use ZFS on root for. My other use-case is to try to use systemd/podman containers on top of zfs volumes to (hopefully) get breezy backups and restores of container data.

> Your message is the first one I received since the guide published. Have you encountered other difficulties during installation?

No actually, after this little blip the installation was smooth sailing. It does boot successfully, but I haven't done anything with it yet since the install. Would you mind if I reach out if I run into problems?

If I had any feedback for the guide, it would be on the Bootloader page. I wasn't sure what I needed to choose between (grub2 vs EFI?), which factors I should use to make the choice, and which steps were needed for each choice. Honestly, out of all the topics covered by this guide my understanding of bootloaders is where I'm the weakest, and I was able to get through it anyway, so feel free to take my feedback here with a grain of salt.

ne9z:
> That's a totally understandable thing to overlook

Thanks for your understanding! Still, there's no excuse to list an ISO, which hasn't been actually tested, as supported.

> Using snapshots to enable rolling back system updates

Despite being mentioned as "rollback" in my previous reply, bieaz actually does something like git branches.

That is, when using DNF, two branches of the system are created and can be selected, at boot, in GRUB menu. With this , it's not needed to overwrite the current system to get back to the previous state.

> use systemd/podman containers on top of zfs volumes to (hopefully) get breezy backups and restores of container data.

A bit of ZFS terminology here. There are two types of filesystems:
- one is called datasets, which presents itself as a mountable filesystem and is used throughout the installation guide
- the other is called zfs volumes, zvol for short, which presents itself as a raw device and has [stability issues](https://github.com/openzfs/openzfs-docs/issues/157#issuecomment-829792424). You most likely wants to avoid it.

One thing of note with boot environments is that, the user needs to do system data/user data separation. i.e., rolling back a faulty kernel update should not affect the virtual machine disk images. Or your documents saved in `/home`.

This means that you need to create separate datasets for the container platform you need to use. For example, for libvirt, we need `zfs create $PATH/DATA/default/var/lib/libvirt`, and so on.

Ideally, the container platform should have an understanding of ZFS, in order to facilitate ZFS-based container snapshot, rollback, etc. podman, being a Red Hat product, most likely lacks native ZFS support and is not recommended for this purpose.

LXD, developed by Canonical and also available on Fedora via a COPR repo, on the other hand, supports ZFS-based container management natively. It will prompt you to set up a separate ZFS dataset during initialization. [Here you can find more details](https://linuxcontainers.org/lxd/getting-started-cli/#linux). Tip: after installation, run `lxd init` and select **do not create separate ZFS pool** and provide a dataset path of your choosing.

> I haven't done anything with it yet since the install.

The guide only installs the CORE package group. In other words, only the packages needed to boot are installed.

Common post installation tasks include installing a desktop environment of your choosing, such as
```
dnf group list
# for gnome
dnf group install 'Fedora Workstation'
```

> Would you mind if I reach out if I run into problems?

Of course not. Just file an issue report at this repo and mention me with @ne9z .

> I wasn't sure what I needed to choose between (grub2 vs EFI?)

GRUB2 supports and is used for both legacy (BIOS) booting and EFI booting.

bieaz also depends on GRUB to work.

> which factors I should use to make the choice, and which steps were needed for each choice.

The choice is mainly whether installing legacy or EFI booting to the hard disk.

However, as mentioned in the guide, you can install both to your hard disk to ensure your computer will boot, regardless of whether legacy or EFI is being used.

EFI Secure Boot is not supported in anyway, due to the presence of a 3rd-party kernel module, ZFS.

ne9z:
Addendum: if you have created separate datasets, say for LXD, in addition to those created during system installation, you need to configure the system to mounted them at boot.

There are three possible ways to mount datasets at boot:

- one is enabling `zfs-mount` service. This is a relatively blunt but most simple way to do the job. It just executes `zfs mount -a` command at boot and can not handle some system directories like `/var`.
- The other one is using `zfs-mount-generator`, described in [post-installation](https://openzfs.github.io/openzfs-docs/Getting%20Started/Fedora/Root%20on%20ZFS/5-bootloader.html#post-installaion), this monitors for dataset creation, property changes and destruction then generates a mounting table in `/etc/zfs/zfs-list.cache`. systemd will read the configuration from here and try to load the encryption keys and mount them at boot.
- The last one is using `/etc/fstab`, used in the installation guide. This is the most granular but most manual way to mount datasets. It also does not handle encryption keys.

The idea of using `/etc/fstab` by default is to achieve multi-distro installation. In fact, you can install Arch Linux, RHEL or NixOS on your Fedora installation right now. All the before-mentioned installation guides are written by myself and configures datasets in the same way.

infogulch:
Thanks for the ZFS terminology correction, I agree that I definitely want to be using datasets not volumes for these uses.

> dnf group list

Yes I saw many interesting package groups there.

Good point about container management engine, I'll look into LXD.

I would like to have each container-volume be its own dataset instead of using a single dataset for all volumes, so I can manage backups and restores on a per-container basis. For example, take a snapshot of a container's dataset on one host, zfs send it to another host, and spin up a container with the same image and cloned dataset on the new host. I expect I may need to wrap this up into a system with some assumptions to make it nice to use, but that's my goal.

ne9z:
> I would like to have each container-volume be its own dataset instead of using a single dataset for all volumes, so I can manage backups and restores on a per-container basis.

That's exactly what LXD does with its ZFS storage backend. Give it a ZFS dataset path, and it will use datasets for everything.

Like: download a vanilla Debian 11 template and unpack it to a new dataset. Creating containers from this template will leverage `zfs clone` function.

Snapshots and rollbacks all uses ZFS specific commands as well, such as `zfs snapshot` and `zfs rollback`.

Of course, the above are all done with `lxc $subcommand`, it's just ZFS is used under the hood. You can see what zfs command got executed with `zpool history $yourpool`.

> For example, take a snapshot of a container's dataset on one host, zfs send it to another host, and spin up a container with the same image and cloned dataset on the new host.

~~My knowledge on send/receiving containers are limited, however I suppose that will involve some LXD config syncing, otherwise LXD on the target computer won't be knowing there's now a new image.~~

Scratch that. LXD natively supports container and image transfer and even zero-downtime live migration. See [LXD introduction](https://linuxcontainers.org/lxd/introduction/), under _Backup and Transfer_ section.

ne9z:
Some examples:

Here's the dataset layout created by LXD, on my NixOS homelab+NAS:
```
rpool/lxd
rpool/lxd/containers
rpool/lxd/containers/jabb # an ejabberd instance
rpool/lxd/custom
rpool/lxd/deleted
rpool/lxd/deleted/containers
rpool/lxd/deleted/custom
rpool/lxd/deleted/images
rpool/lxd/deleted/virtual-machines
rpool/lxd/images
rpool/lxd/virtual-machines
```

View what LXD did when I created the `jabb` container:
```
zpool history rpool
zfs create -o mountpoint=/var/lib/lxd/storage-pools/default/images/Deb11TemplateHash -o canmount=noauto rpool/lxd/images/zfs set quota=none rpool/lxd/images/Deb11TemplateHash
zfs snapshot rpool/lxd/images/Deb11TemplateHash@readonly
zfs clone rpool/lxd/images/Deb11TemplateHash@readonly rpool/lxd/containers/jabb
zfs set mountpoint=/var/lib/lxd/storage-pools/default/containers/jabb canmount=noauto rpool/lxd/containers/jabb
zfs set quota=none rpool/lxd/containers/jabb
```

ne9z:
Another addendum:

if you are using ZFS encryption, then you must carefully choose the LXD dataset path.

A bit about ZFS native encryption: this encryption is enabled on a per-dataset basis, and child datasets nested in a parent dataset by default inherits the encryption properties of the parent, that is, same decryption password. This relation can also be broken later.

The parent dataset is the encryption root of the child dataset. You can view the encryption root of your system with this command `zfs list -o name,encryption,encroot`.

In the guide, the encryption root is `rpool/fedora`. If LXD is nested inside this dataset, the containers then will be encrypted with the same passphrase as the system. You can also choose to create another encryption root specifically for LXD, with a different passphrase or keyfile.

ZFS encryption key is a master key wrapped with the supplied passphrase.
When changing passphrase with `zfs change-key`, only the supplied passphrase is changed, the master key stays the same. This is necessary for the Copy-on-Write model ZFS is using.

Generally speaking, there are both advantages and drawbacks of the per-dataset encryption compared to per-disk encryption like LUKS.

Advantages:
- You can safely send an encrypted dataset with the raw option `zfs send -w`, without key, to a remote computer and let it do the scrubbing for you
- More flexibility, like separate encryption roots, separate passphrases and so on.
- Other things I don't know

Drawbacks:
- Metadata like dataset name can be viewed without decryption.
- Due to the copy-on-write nature of ZFS, once the passphrase is compromised, on-disk content can not be reencrypted with a new passphrase, you are stuck with the same master key forever unless you destroy and recreate the dataset. Previously encrypted content are likely to be recovered with the old passphrase with enough effort.
- Other things I don't know

Of course I'm no expert on this matter, only a normal user who hasn't even touched one line of ZFS code, so take my advice with a grain of salt and carefully evaluate your own security concerns.

gmelikov:
@ne9z could you please approve it if everything is ok here for you?

ne9z (approved):

ne9z:
My fault. Approved.

gmelikov (approved):

PR #209, opened 2021-10-02
Update FreeBSD install instructions

allanjude:
Default version if python 3.8 now, and remove an excess \


gmelikov (approved):
Thank you!

PR #211, opened 2021-10-09
add `keymap` hook to initrd

j5k:
Add `keymap` hook to initrd for correct keyboard layout on decryption password input.

https://wiki.archlinux.org/title/Mkinitcpio#Common_hooks


ne9z:
That's a valid point. Add a explicit note for keymap and another link
for other hooks then.

j5k:
You can do what you want. It would help all non-US layout user.

PR #214, opened 2021-10-13
Use an environment variable for username in step 6.6

scott-ainsworth:
In step 6.6, Create a user account, use an environment variable for the user name to reduce the possibility of typos and simplify using copy-and-paste to run commands.

Signed-off-by: Scott G. Ainsworth <scott@ainsworth.us>

### DIFF
--- a/docs/Getting Started/Ubuntu/Ubuntu 20.04 Root on ZFS.rst
+++ b/docs/Getting Started/Ubuntu/Ubuntu 20.04 Root on ZFS.rst
@@ -1071,17 +1071,18 @@ Step 6: First Boot
 
    Replace ``username`` with your desired username::
 
+     UserName=username
      UUID=$(dd if=/dev/urandom bs=1 count=100 2>/dev/null |
          tr -dc 'a-z0-9' | cut -c-6)
      ROOT_DS=$(zfs list -o name | awk '/ROOT\/ubuntu_/{print $1;exit}')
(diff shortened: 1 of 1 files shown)


scott-ainsworth:
If you would like me to create a new branch and consolidate the three commits into one, please let me know.

rlaager:
> If you would like me to create a new branch and consolidate the three commits into one, please let me know.

It doesn't need to be a new branch, but ideally you would squash these commits into one. You can do that with something like `git rebase -i origin/master`, leave the first marked as "pick", and mark the others as "squash" (or "s") to squash them into the one above. This will give you an editor to change the commit message. In this case, you would be discarding the subsequent commit messages anyway, so instead of "squash" you can use "fixup" (or "f") in the rebase. Then do a _force_ push, e.g.: `git push -f`

If that rebasing/squashing is too confusing for you and you'd rather not, I can do it as part of accepting this PR if you'd prefer.

scott-ainsworth:
Rebase/squash is old hat. Just wanted to check since not everyone likes it.  Just pushed the (hopefully) final version.

Thank you for your patience!

Scott

rlaager:
I renamed it to $username, as that is what I would prefer. I applied the change to the other HOWTOs (Ubuntu 20.04 for Raspberry Pi and Debian Buster--I still haven't done a Bullseye one). Merged as 3c04e929a1ed86430a924485adda38fcfd5d776e.

Thanks!

Issue #215, opened 2021-10-13
Questions about Debian initrd documentation


Yvan-Masson:
Hi,

Thanks for documenting OpenZFS. I am experimenting on a fresh Bullseye VM with root on ZFS, following the current PR #205. Grub is installed in BIOS mode and there is only one disk. System works properly, but the current [Debian initrd documentation](https://openzfs.github.io/openzfs-docs/Getting%20Started/Debian/Debian%20GNU%20Linux%20initrd%20documentation.html) does not match what I have:

1. The Grub linux command line in the "Booting from snapshot" section is different. If we remove the snapshot mentioned to have the original line, the doc gives :
```
linux   /ROOT/debian-1@/boot/vmlinuz-3.2.0-4-amd64 root=ZFS=rpool/ROOT/debian-1 ro boot=zfs $bootfs quiet
```
While on my system it is:
```
linux   /BOOT/debian@/vmlinuz-5.10.0-9-amd64 root=ZFS=/ROOT/debian ro root=ZFS=rpool/ROOT/debian quiet
```
Maybe it is just Grub that is much newer in Bullseye than in the documentation?

2. The doc says it should be possible to select snapshot dynamically by adding `@` only to the `root` parameter. In my case, system just hangs after `Loading initial ramdisk ...`

3. I tried a rollback from snapshot by specifying the snapshot name in the `root` parameter and by adding `rollback=1`. System then hangs (see picture below). After enforcing a reset, system boot properly on the previously selected snapshot. Maybe this could be better explained in the documentation?
![snapshot_rollback](https://user-images.githubusercontent.com/9531149/137141894-0e01820f-71f4-40f4-9c15-b83d36fc63fd.png)

Regards,
Yvan

rlaager:
The Root on ZFS HOWTOs use a separate bpool and have for a while now. I didn't write those other docs. I've made some minor tweaks to bring it closer.

PR #218, opened 2021-10-19
Root on ZFS Raspberry Pi for USB SSDs

scott-ainsworth:
Modify the `Ubuntu 20.04 Root on ZFS for Raspberry Pi` instructions to allow installation on a USB SSD (or HDD, flash drive, etc.) To make the USB device bootable, the Linux image must be decompressed. Work by [James Chambers](https://jamesachambers.com/raspberry-pi-4-ubuntu-20-04-usb-mass-storage-boot-guide/) is incorporated to accomplish this. The three commits separate two set of ancillary changes from the main work.

### DIFF
--- a/docs/Getting Started/Ubuntu/Ubuntu 20.04 Root on ZFS for Raspberry Pi.rst
+++ b/docs/Getting Started/Ubuntu/Ubuntu 20.04 Root on ZFS for Raspberry Pi.rst
@@ -12,7 +12,7 @@ Overview
 Caution
 ~~~~~~~
 
-- This HOWTO uses a whole physical SD card.
+- This HOWTO uses a whole physical disk.
 - Backup your data. Any existing data will be lost.
 
 System Requirements
@@ -22,11 +22,13 @@ System Requirements
   :doc:`Ubuntu 20.04 Root on ZFS`.)
 - `Ubuntu Server 20.04.3 (“Focal”) for Raspberry Pi 4
   <https://cdimage.ubuntu.com/releases/20.04.3/release/ubuntu-20.04.3-preinstalled-server-arm64+raspi.img.xz>`__
-- A microSD card or USB disk. For microSD card recommendations, see Jeff Geerling's `performance
-  comparison
+- A microSD card or USB disk. For microSD card recommendations, see Jeff
+  Geerling's `performance comparison
   <https://www.jeffgeerling.com/blog/2019/raspberry-pi-microsd-card-performance-comparison-2019>`__.
-- An Ubuntu system (with the ability to write to the SD card or USB disk) other than the
-  target Raspberry Pi.
+  When using a USB enclosure, `ensure it supports UASP
+  <https://github.com/geerlingguy/turing-pi-cluster/issues/11#issuecomment-647726561>`__.
+- An Ubuntu system (with the ability to write to the microSD card or USB disk)
+  other than the target Raspberry Pi.
 
 4 GiB of memory is recommended. Do not use deduplication, as it needs `massive
 amounts of RAM <http://wiki.freebsd.org/ZFSTuningGuide#Deduplication>`__.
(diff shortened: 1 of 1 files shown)


rlaager:
I don't fully understand why we need to decompress the kernel. The linked HOWTO says it's because we get rid of u-boot. Is that a thing we need to do?

scott-ainsworth:
Forced pushed a new set of commits a few minute ago. These incorporate the change requests and separate the primary changes from non-essential changes.

Thanks,
Scott

scott-ainsworth:
I have encountered an interesting issue with the `/etc/kernel/postinit.d` hook. I needs to run after `zz-flash-kernel`, so I named the decompression script `zzz-decompress-kernel`. However, `zz-flash-kernel` defers its updates; thus, the kernel has not yet been updated when `zzz-decompress-kernel` is run. I'm now looking at how deferral works.

scott-ainsworth:
So, `zz-flash-kernel` runs after the kernel is in place and after `xx-update-initrd-links` updates the symlinks. It looks like `zzz-decompress-kernel` can use `/boot/vmlinuz` instead of `/boot/firmware/vmlinuz`.

rlaager:
I tested this. I merged a bunch of stuff (to master) I found along the way, a change of yours, and some stuff inspired by the changes here.

I've pushed (to your branch) a rebased version of your changes plus my edits. Some are trivial (like rewrapping), some are intermediate (like some more links or wording changes), and some are significant. Can you take a look at my commit here (either that individual commit or by looking at the whole net change of the PR) and see if you see any issues?

Some specific things:

Is `dtoverlay=vc4-fkms-v3d` actually related to this in some way or is that just something that slipped in?

I think we need to write to usercfg.txt, not config.txt these days.

In thinking more about this, we should not rely on timestamps, since the Pi has no real-time clock. And in thinking even more, I don't see any reason we need to check for changes. If an /etc/kernel/postinst.d script is running, it's because a kernel was installed. We can just decompress unconditionally. Checking made more sense back when this was an apt dpkg hook that ran all the time for unrelated reasons.

scott-ainsworth:
I reviewed the individual commits and tested the entire process from scratch (except for encryption). 

The code block for 01-netcfg.yaml (Step 6.1) has an extra leading space. So, when the block is copy-pasted into vi, the lines are shifted right 1 column (e.g., "network:" starts in column 2 vice column 1). Commit 58bdea2 removes the extra leading space. 

The `vc4-fkms-v3d` overlay enables 3D graphics acceleration. For a server install, it is not needed. For a desktop system, it is beneficial. However, to be effective, additional information must included in `usercfg.txt`. Instead of holding this pull request up, I have written issue #224 to follow up.

The unattended upgrade interference I thought I solved with `systemctl stop unattended-upgrade` raised its head again. It turns out that stopping the service is only effective done shortly after reboot (< 1 minute?). After that a kernel upgrade might be in progress and the service stops only after the kernel upgrade is complete. After the upgrade finishes, installation can continue. However, stopping the unattended kernel upgrade can render the media unbootable (because `/etc/kernetl/postinit.d/zz-deccompress-kernel` might not have run). Unbootable media and user confusion can be prevented two ways: (a) include a note about waiting for the unattended upgrade to complete or (b) turn it off in Step 1 and back on in Step 7. This might apply to more than just the Raspberry Pi and seems to warrant additional discussion. Issue #215 addresses these questions.

`addgroup --system lxd` produced a warning because group lxd was already present. Since I copy-pasted all three lines, the warning prevented the next line (add group sambashare) from running. The `--quiet` options suppresses this warning. Commit dd8d32a adds the quiet option.

rlaager:
Merged manually as f60e80083d96b948baad711f6faa71001db2b12c. This includes my changes here and a couple more.

Thank you for this awesome contribution!

scott-ainsworth:
You are very welcome!

PR #223, opened 2021-10-29
wrong logical order

JOduMonT:
should exit the chroot, umount pool and reboot before talking about the 1st boot ;)


rlaager:
I take your point, but I'm generally pretty happy with how this is.

I made some changes in 41ea6da337d6d458fabe8e432f548630eaeb0bee. I didn't move everything up, as I don't want those steps in the section titled "GRUB Installation" either. So really I just moved up the "Install SSH" step, as that's not directly related to the first reboot process.

JOduMonT:
@rlaager ;) hopefully you have more experience with this procedure than me, the main thing for me that, it was odd to ask to quit the chroot environment and unmount then reboot after the First Boot :)

PR #229, opened 2021-11-05
Correct UUID v4 generation for unique pool suffix

pmarreck:
Original example is wasteful and doesn't even produce a UUID as the variable name its result is assigned to would indicate- it uses the alphabet a-z0-9 and not a-f0-9.
Replacement produces an actual UUIDv4 (16 bytes or 32 hex chars), does it more directly, and uses random instead of urandom, which is more secure, and is unlikely to block in this case due to the more concise way it's computed.
While it introduces a new dependency, the xxd hexdump utility is present on every Arch install I've seen thus far.

### DIFF
--- a/docs/Getting Started/Arch Linux/Root on ZFS/1-preparation.rst
+++ b/docs/Getting Started/Arch Linux/Root on ZFS/1-preparation.rst
@@ -60,7 +60,7 @@ Preparation
    unique, therefore it's recommended to create
    pools with a unique suffix::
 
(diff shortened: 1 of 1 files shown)


rlaager:
From the Ubuntu side (from which this derives), despite being called a "UUID", this was never intended to be a real UUID. It is indeed "wasteful" but it's done that way in the initramfs due to limited tooling, and it's done that way elsewhere for compatibility.

I'm going to close this. If @ne9z feels differently, it can be reopened.

PR #230, opened 2021-11-21
Fix /boot mounting and add notice for other keyboard layouts

ne9z:
- Fix NixOS /boot mounting: Add mount option to automatically recreate mountpoint
- Add notice for other keyboard layouts

Closes #211 

@gmelikov


gmelikov (approved):
Thank you!

PR #231, opened 2021-11-21
GRUB: preliminary support for PBKDF2-only LUKS2

ne9z:
Affects: NixOS and Arch Linux Root on ZFS guides, encrypted boot pool section.

In GRUB 2.06, support for PBKDF2-only LUKS2 is introduced. Compared to LUKS1, currently used by the guides, LUKS2 offers a series of advantages. A list can be found on [development wiki](https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions#10-luks2-questions)

Currently I have confirmed that, when `luks2 pdkf2 cryptodisk` modules are loaded, GRUB 2.06 can indeed open a PBKDF2-only LUKS2 container.

Whether `grub-mkconfig` can automatically detect LUKS2 container then insert necessary modules to `grub.cfg` during bootloader installation, is not tested.

### DIFF
--- a/docs/Getting Started/Arch Linux/Root on ZFS/4-optional-configuration.rst
+++ b/docs/Getting Started/Arch Linux/Root on ZFS/4-optional-configuration.rst
@@ -197,6 +197,7 @@ root pool will be replaced by keyfile, embedded in initrd.
 #. Let GRUB decrypt all LUKS containers on boot::
 
      tee -a /etc/grub.d/09_bpool_luks2-decryption <<FOE
+     #!/bin/sh
      cat <<EOF
        insmod luks2
        insmod pbkdf2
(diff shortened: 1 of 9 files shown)


gmelikov (approved):

gmelikov:
@ne9z ah, it's marked as Draft, so waiting when it would be ready.

ne9z:
In the current form, where GRUB can not properly detect LUKS2 and add menu entries for the encrypted container, support on NixOS seems to be impossible. So that part was reverted back to LUKS1.

On the other hand Arch Linux is now able to boot from a LUKS2 encrypted boot pool.

ne9z:
@gmelikov Ready for merge.

PR #238, opened 2021-12-14
one liner install command does not work

p4tpr0:
the one-liner "dnf install epel-release kernel-devel zfs" does not work as intended, at least on RockyLinux 8.5. You need to break the line in 2 commands: one for epel and one for other two packages.

### DIFF
--- a/docs/Getting Started/RHEL-based distro/index.rst
+++ b/docs/Getting Started/RHEL-based distro/index.rst
@@ -68,7 +68,8 @@ For RHEL/CentOS versions 6 and 7 run::
 
 And for RHEL/CentOS 8 and newer::
 
- dnf install epel-release kernel-devel zfs
+ dnf install epel-release
+ dnf install kernel-devel zfs
 
 .. note::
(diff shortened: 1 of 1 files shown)


rlaager (approved):
I see no problem with this, but I can't say I understand the issue either. I certainly have not tested. So I'm not going to merge this, but I did give it a review in case someone else wants that.

p4tpr0:
Problem is, one apparently can't install the 3 packages in one command, it yields to this situation: https://twitter.com/p4tpr0/status/1469732134990618633
Installing epel-release in first command, then installing kernel-devel zfs in a second command is the only way to get it done.

rlaager:
Right, I'm not disagreeing. I was just qualifying my review.

ne9z (approved):

ne9z:
@rlaager I wrote this part and can confirm this bug. I think it is ready for merging.

gmelikov (approved):

gmelikov:
Thank you! Fixed as part of #239

PR #243, opened 2021-12-23
RHEL: switch to kmod

ne9z:
Switch to kmod can save time and electricity from compilation. Currently, in order to perform a Rocky Linux Root on ZFS installation, DKMS package needs to be built for both live and installed system.

However, sometimes the latest version of kernel in Rocky repo might be incompatible with kmod package in zfsonlinux repo. In other words, zfsonlinux repo might lag behind official kernel releases, causing the installation to fail.

Request a review from @tonyhutter 

### DIFF
--- a/docs/Getting Started/RHEL-based distro/RHEL 8-based distro Root on ZFS/1-preparation.rst
+++ b/docs/Getting Started/RHEL-based distro/RHEL 8-based distro Root on ZFS/1-preparation.rst
@@ -43,6 +43,8 @@ Preparation
 #. Install ZFS packages::
 
     dnf install -y epel-release
+    dnf config-manager --disable zfs
+    dnf config-manager --enable zfs-kmod
     dnf install -y zfs
 
 #. Load kernel modules::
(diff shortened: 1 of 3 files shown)


gmelikov:
> However, sometimes the latest version of kernel in Rocky repo might be incompatible with kmod package in zfsonlinux repo. In other words, zfsonlinux repo might lag behind official kernel releases, causing the installation to fail.

So, it's better to use DKMS, no?

ne9z:
It depends on the hardware, I think.

For a computer with tons of computing power and memory the benefit of kmod will be negligible for sure.

But for a computer with limited processing power and memory, such as a NAS with Intel Pentium and 2GB of ECC RAM, two times of DKMS compilation will not only use a lot of time, it might even fail due to the lack of memory.

For me, who tests the Root on ZFS installation guide inside a dual core, 2GB RAM virtual machine, kmod has consistently proved itself to be a lifesaver, whereas DKMS will simply exhaust the memory and fail.

Furthermore, from my personal experience, DKMS compilation might also silently fail without the user ever noticing. If this is the case, the user could end up with an on-disk system without ZFS support, leading to a kernel panic on first boot. With kmod, if the kernel is compatible, the installation is guaranteed to be successful.

Therefore, if we could keep kmod in sync with official kernel updates, I think it would be beneficial to switch to kmod by default.

gmelikov (approved):

mskarbek:
IMO DKMS should be treated as a secondary option in most/every scenario. It is useful but over the years also proved that it is unreliable and a source on many issues and users frustration.

PR #244, opened 2021-12-29
Update errata to remove reference to old paths

shotcollin:
This pull request fixes (what I think may be) two erroneous path references in the errata section related to the `bpool` GRUB dataset.


rlaager:
I think this change is correct. I merged it. Thanks!

PR #247, opened 2021-12-30
Update OpenZFS Exceptions.rst

jumbi77:


gmelikov (approved):
We have these exceptions only for history, but I'm ok to append some info while it's already here, thank you.

PR #253, opened 2022-01-14
zfs doc: upper bound for slop space calculation

jumbi77:
OpenZFS v2.1 and above added an upper bound for slop space calculation.
See: https://github.com/openzfs/zfs/commit/f01eaed4556623b63c414fee9085ae27d457fe46


gmelikov (approved):
Thank you!

PR #254, opened 2022-01-16
Use 'private' bind mounts

kpfleming:
When plain '--rbind' is used, the 'umount' process executed later in the
setup process completely unmounts the bound filesystem, resulting in various
forms of breakage on the host used to do the setup. When a Live CD is used
and the system will be immediately shutdown/rebooted, this isn't an issue,
but in other scenarios it can be problematic.

Making the bind mounts 'private' avoids this problem. This patch changes
the setup guides for 'modern' distribution versions but does not touch
the older ones where this functionality may not be available.

Signed-off-by: Kevin P. Fleming <kevin@km6g.us>

CC: @rlaager

### DIFF
--- a/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
@@ -584,9 +584,9 @@ Step 4: System Configuration
 #. Bind the virtual filesystems from the LiveCD environment to the new
    system and ``chroot`` into it::
 
-     mount --rbind /dev  /mnt/dev
-     mount --rbind /proc /mnt/proc
-     mount --rbind /sys  /mnt/sys
+     mount --make-private --rbind /dev  /mnt/dev
+     mount --make-private --rbind /proc /mnt/proc
+     mount --make-private --rbind /sys  /mnt/sys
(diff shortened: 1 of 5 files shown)


rlaager:
> When plain '--rbind' is used, the 'umount' process executed later in the
setup process completely unmounts the bound filesystem

Can you provide some instructions on how to reproduce this?

Are you saying that if I `mount --rbind /dev /mnt/dev` and then `umount` the various `/mnt/dev/*`, that I will ultimately unmount `/dev` too? I am not able to reproduce this with Debian 11.

kpfleming:
Sorry, I should have provided more specifics.

On a Debian 11 system, `mount --rbind /dev /mnt/dev` followed by the loop-unmount sequence from the guide results in `/dev/pts`, `/dev/shm`, and the various other filesystems *below* the explicitly-mounted ones to be unmounted. I noticed the problem after trying to open an SSH connection into the host system after completing the process, and was unable to do so because `/dev/pts` was no longer mounted.

rlaager:
With that information, I was able to reproduce this. That's really interesting (and feels like a bug to me). Your proposed `--make-private` change does indeed fix it. So I merged this as 5713bc08ed108275c26dbbdd61a5492583feeade (so I could add my Signed-off-by). Thanks!

PR #260, opened 2022-01-31
Arch Linux: Update LiveISO. 2022.01.01

ne9z:
2022.01.01: linux 5.15.12
zfs: 2.1.2; supported

@gmelikov We need to either automate updating live iso links or find another people as collaborator: I might not be around to update this link.

The procedure is documented [here](https://openzfs.github.io/openzfs-docs/Getting%20Started/Arch%20Linux/3-live.html).

### DIFF
--- a/docs/Getting Started/Arch Linux/Root on ZFS/1-preparation.rst
+++ b/docs/Getting Started/Arch Linux/Root on ZFS/1-preparation.rst
@@ -7,8 +7,8 @@ Preparation
    :local:
 
 #. Download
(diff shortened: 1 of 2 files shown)


ne9z:
Side note: Arch Linux Live ISO is good for only 3 months. Afterwards it will be deleted from server.

Reference: https://archlinux.org/releng/releases/

gmelikov:
@ne9z Let's just give https://mirrors.ocf.berkeley.edu/archlinux/iso/latest/ and users can find latest iso by themselves.

ne9z:
I really hope we can do just that.  But sometimes the latest live media might contain a kernel version incompatible with ZFS.

ne9z:
I will create a new PR to let user find the compatible version themselves.

PR #262, opened 2022-02-01
 Workaround solution with a package

Rudd-O:
This PR add instructions to use a package to apply the workaround for GRUB.

### DIFF
--- a/docs/Getting Started/Fedora/Root on ZFS/5-bootloader.rst
+++ b/docs/Getting Started/Fedora/Root on ZFS/5-bootloader.rst
@@ -29,12 +29,18 @@ Workarounds have to be applied.
 
    See `this bug report <https://savannah.gnu.org/bugs/?59614>`__.
    Root pool name is missing from ``root=ZFS=rpool_$INST_UUID/ROOT/default``
-   kernel cmdline in generated ``grub.cfg`` file.
+   kernel cmdline in generated ``grub.cfg`` file.  This can also manifest as
+   an error message `cannot find a device for /` during `grub2-mkconfig`.
 
-   A workaround is to replace the pool name detection with ``zdb``
-   command::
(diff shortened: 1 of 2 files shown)


ne9z:
Thanks for the contribution!

IMHO this addition adds confusion to the installation guide and should not be accepted:

- The guide is not a collection of alternatives: only the most simple, straightforward and effective way is kept.
- The guide is intended to be executed in a sequential manner and all instructions need to be present: the user must be able to rely on this guide alone to successfully install the system.  Therefore no "google how to build the package and install it yourself" section.
- If it is a Fedora/RHEL package, it should be built and distributed by COPR. Or else future updates can not be done easily.

A few questions:
- Does this package have any advantage over the current one-liner solution? Note that the one liner is also included in the boot environment manager bieaz, see [here](https://gitlab.com/m_zhou/bieaz/-/blob/0.3.3/bieaz#L216).
- If it does, can you publish the package on COPR and test against past and future Fedora/RHEL revisions?
- This guide is authored in reStructuredText format, it appears your additions are written in Markdown.

Rudd-O:
The problem with the guide as-is, is that an upgrade to the GRUB package on a system that follows the guide will render the next kernel update unbootable for all kernels.

The proposed package fixes that permanently on the system.

I regret not being able to invest more effort in this than the one I have already invested.  Sorry.

ne9z:
No regrets needed ;-)

> upgrade to the GRUB package on a system that follows the guide will render the next kernel update unbootable for all kernels

I think this is solved by installing the boot environment manager.  That package contains the fix and, with the help of the dnf hook, will execute the one-liner fix whenever some package is upgraded.

I had a quick look in your package, and it seems to me that the one-liner fix is needed anyway because:

> You have run grub2-mkconfig at least once.

Notice at this stage of installation, this command haven't been run yet.

ne9z:
Can you close this pull request?  Your instructions is now out-of-date with the latest Root on ZFS guide.

PR #265, opened 2022-02-06
NixOS: remove encrypted swap section due to system bug

ne9z:
See https://github.com/NixOS/nixpkgs/issues/157989

So we will temporarily remove this section to prevent boot failures.

@gmelikov 


gmelikov:
@ne9z I'll merge it as is, after https://github.com/NixOS/nixpkgs/issues/157989 is fixed please make it as a commit revert.

PR #271, opened 2022-02-19
Add kernel module installation to Ubuntu manual

szubersk:
Dumb down the description to help new users install ZFS:
- mention `zfs-dkms` package requirement
- mention `universe` component in `apt.sources`
- add post-installation checks to confirm that ZFS was installed
  correctly

Tested on Debian bookworm host using debootstrap following releases:
- bionic
- focal
- impish
- jammy

Fix openzfs/zfs#13061

Signed-off-by: szubersk <szuberskidamian@gmail.com>

### DIFF
--- a/docs/Getting Started/Ubuntu/index.rst
+++ b/docs/Getting Started/Ubuntu/index.rst
@@ -7,14 +7,40 @@ Ubuntu
 Installation
 ------------
 
-If you want to use ZFS as your root filesystem, see the `Root on ZFS`_
-links below instead.
+.. note::
+  If you want to use ZFS as your root filesystem, see the
+  `Root on ZFS`_ links below instead.
 
-Since Ubuntu 16.04, ZFS is included in the default Linux kernel packages.  To
-install the ZFS utilities::
+.. note::
+  Steps in this guide apply to most Ubuntu-based distributions.
+
+Since Ubuntu 16.04 ("Xenial Xerus"), ZFS is included in the default
+Linux kernel packages. Due to license issues, ZFS kernel modules
+must be built from sources by dkms.
+
+To install the ZFS first make sure ``universe`` component is enabled
+in ``/etc/apt/sources.list``:
+
+::
+
+  deb http://archive.ubuntu.com/ubuntu <UBUNTU CODE NAME> main universe
+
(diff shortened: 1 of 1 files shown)


rlaager:
Thanks for the patch!

I manually merged this in part as e004a424e4940b06edf359a41af114a337ed9512 and 77b8563015aafe3029d644a22ff342481bebb16f. In the middle (96c790d51bfa299321f8c3f3b4c96f78c6dc0b87), I dropped all mentions of 16.04, because it's been EOL for almost a year now.

Specifically, I kept the format change to the note and the mention of adding universe. I can see arguments for the post-installation checks, but the user is going to find the same answer as soon as they run _any_ zfs/zpool command, so ultimately I dropped that.

`zfs-dkms` is _not_ required on Ubuntu. I understand that some Ubuntu derivatives handle this differently. That's their choice, but I'm not trying to cover those scenarios in Ubuntu instructions any more than I'm trying to cover Ubuntu scenarios in Debian instructions.

szubersk:
> `zfs-dkms` is _not_ required on Ubuntu. I understand that some Ubuntu derivatives handle this differently. That's their choice, but I'm not trying to cover those scenarios in Ubuntu instructions any more than I'm trying to cover Ubuntu scenarios in Debian instructions.

The empirical proof at https://github.com/openzfs/zfs/issues/13061#issuecomment-1036463367. Missing `zfs-dkms` was the root cause of the issue. The fact that issue was reported originally for PopOS is irrelevant, exactly the same issue shows up on Ubuntu for the same reason.

gmelikov:
@szubersk but in Ubuntu itself https://packages.ubuntu.com/search?keywords=zfs-dkms it's `also provided by` every kernel, so module is shipped with kernel, don't see an issue with Ubuntu here, what do I miss?

szubersk:
> @szubersk but in Ubuntu itself https://packages.ubuntu.com/search?keywords=zfs-dkms it's `also provided by` every kernel, so module is shipped with kernel, don't see an issue with Ubuntu here, what do I miss?

Apologize, looks like I missed the point :|
```
doas debootstrap --no-check-gpg  --include=zfsutils-linux,linux-image-generic focal ./chroot
```
works perfectly fine. Indeed, Ubuntu comes with precompiled module.

Issue #270, opened 2022-02-17
Unable to export rpool before first reboot


reefland:
@rlaager, I have the following issue with the Ubuntu 20.04 Root on ZFS HOWTO:

I've been using the ZFS on Root for Ubuntu for several years since 18.04 LTS.  I've even replicated much of it as ansible playbook.  I completed step updates based on guide to work with 20.04, then 20.04.1, now I'm testing with 20.04.3 and having a problem.

Following: https://openzfs.github.io/openzfs-docs/Getting%20Started/Ubuntu/Ubuntu%2020.04%20Root%20on%20ZFS.html#overview

I use a VirtualBox environment to practice / refine the steps before attempting on bare metal.  I'm testing with three 20GB VDI files which provide for a 3 disk Legacy Boot, ZFS Native Encryption, Raidz1 setup, as per guide mdm is being used for encrypted swap. 

VirtualBox boots the Ubuntu Desktop 20.04.3 ISO image and I follow the steps provided in the guide.   I have started from scratch several times and keep having the same issue:

I get to Section 5, GRUB Installation, Step 10 ... I'm unable to export the pools.  Everything previous to this goes pretty smooth.

My troubleshooting trying to determine why the pool busy:

I'm not in a path used by rpool:
```
# pwd
/home/ansible
```
Everything appears to unmounted cleanly:
```
# lsof /mnt
lsof: WARNING: can't stat() fuse.gvfsd-fuse file system /run/user/999/gvfs
      Output information may be incomplete.

# lsof | grep "/mnt"   
lsof: WARNING: can't stat() fuse.gvfsd-fuse file system /run/user/999/gvfs
      Output information may be incomplete.
    
# ls -l /mnt
total 0

# mount | grep -v zfs | tac | awk '/\/mnt/ {print $3}'
```
No swap is used at this point:
```
# cat /proc/swaps
Filename                                Type            Size            Used            Priority

# free -mh
              total        used        free      shared  buff/cache   available
Mem:          7.8Gi       2.6Gi       3.2Gi       425Mi       2.0Gi       4.5Gi
Swap:            0B          0B          0B
```
The pool itself is in good shape:
```
# zpool list
NAME    SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
rpool    53G  1.56G  51.4G        -         -     0%     2%  1.00x    ONLINE  /mnt

# zpool status -v
  pool: rpool
[...]

PR #289, opened 2022-03-30
bugfix/removing unnecessary efi mount/umount commands

djacu:
This fixes an issue where the OS would through an error during the system installation. There was a `umount` call that didn't unmount anything and would produce an error as shown:
```sh
$ umount /mnt/boot/efi
umount: /mnt/boot/efi: no mount point specified
```

There were also lines in the zfs.nix file under `grub.extraPrepareConfig` that were removed because `/boot/efi` is not used but `/boot/efis` is.

<strike>
This fixes an issue where the file system was not unmounted completely before exporting the zpools.

There was a `umount` call previously that didn't unmount anything and would produce an error as shown:
```sh
$ umount /mnt/boot/efi
umount: /mnt/boot/efi: no mount point specified
```

Additionally, the newly created zpools were not completely unmounted and would cause them to not be imported properly after reboot. I believe this is related to #270 as the symptoms are very similar, but I cannot confirm as I have not tried the same installation.
</strike>

### DIFF
--- a/docs/Getting Started/NixOS/Root on ZFS/2-system-configuration.rst
+++ b/docs/Getting Started/NixOS/Root on ZFS/2-system-configuration.rst
@@ -333,9 +333,8 @@ System Configuration
         grub.zfsSupport = true;
         # for systemd-autofs
         grub.extraPrepareConfig = ''
-          mkdir -p /boot/efis /boot/efi
+          mkdir -p /boot/efis
           for i in  /boot/efis/*; do mount $i ; done
-          mount /boot/efi
         '';
         grub.extraInstallCommands = ''
            export ESP_MIRROR=$(mktemp -d -p /tmp)
--- a/docs/Getting Started/NixOS/Root on ZFS/3-optional-configuration.rst
+++ b/docs/Getting Started/NixOS/Root on ZFS/3-optional-configuration.rst
@@ -78,7 +78,6 @@ root pool will be replaced by keyfile, embedded in initrd.
     for i in ${DISK}; do
      umount /mnt/boot/efis/${i##*/}-part1
     done
-    umount /mnt/boot/efi
 
 #. Destroy boot pool::
 
@@ -160,8 +159,6 @@ root pool will be replaced by keyfile, embedded in initrd.
     for i in ${DISK}; do
      mount ${i}-part1 /mnt/boot/efis/${i##*/}-part1
     done
-
-    mount -t vfat ${INST_PRIMARY_DISK}-part1 /mnt/boot/efi
 
 #. As keys are stored in initrd,
    set secure permissions for ``/boot``::
--- a/docs/Getting Started/NixOS/Root on ZFS/4-system-installation.rst
+++ b/docs/Getting Started/NixOS/Root on ZFS/4-system-installation.rst
@@ -86,7 +86,6 @@ Finish installation
 #. Unmount EFI system partition::
 
     umount /mnt/boot/efis/*
-    umount /mnt/boot/efi
 
 #. Export pools::
 


ne9z:
The `umount /mnt/boot/efi` line is actually a left-over from other guides and should be completely removed.  In the NixOS guide, `/mnt/boot/efi` is not used at all.

> the newly created zpools were not completely unmounted and would cause them to not be imported properly after reboot

They will be completely unmounted during `zpool export` step, not before it.  The export step will automatically unmount all zfs filesystems.

It can not take care of anything non zfs, so we need to unmount EFI system partitions first.

> would cause them to not be imported properly after reboot

I can not reproduce this issue.  Please ensure that all pools are correctly exported before reboot.

djacu:
@ne9z are you saying in your system you only have `umount /mnt/boot/efis/*` and no other `umount` commands?

When I tried running it from a script, the `zpool export` command would fail but was not printing anything to the console. When I run it manually, it failed and did say something about the pool still being busy. I'll run it again and get the exact message when I leave off the `umount -R /mnt` line.

djacu:
Hey @ne9z I have narrowed down the issue I was having. I believe it has something to do with the area where datasets were created. I modified your script to be more like mine until it failed in the same way that my script originally did.

Here is the dataset creation in your script. It has some slight formatting modifications for readability but otherwise works exactly the same as the file you linked previously.
```bash
zfs create -o canmount=off -o mountpoint=none bpool_$INST_UUID/$INST_ID
zfs create -o canmount=off -o mountpoint=none bpool_$INST_UUID/$INST_ID/BOOT
zfs create -o canmount=off -o mountpoint=none rpool_$INST_UUID/$INST_ID/ROOT
zfs create -o canmount=off -o mountpoint=none rpool_$INST_UUID/$INST_ID/DATA
zfs create -o mountpoint=/boot -o canmount=noauto bpool_$INST_UUID/$INST_ID/BOOT/default
zfs create -o mountpoint=/ -o canmount=off rpool_$INST_UUID/$INST_ID/DATA/default
zfs create -o mountpoint=/ -o canmount=off rpool_$INST_UUID/$INST_ID/DATA/local
zfs create -o mountpoint=/ -o canmount=noauto rpool_$INST_UUID/$INST_ID/ROOT/default
zfs mount rpool_$INST_UUID/$INST_ID/ROOT/default
zfs mount bpool_$INST_UUID/$INST_ID/BOOT/default

for i in {nix,}; do
    zfs create -o canmount=on -o mountpoint=/$i rpool_$INST_UUID/$INST_ID/DATA/local/$i;
done

zfs create -o canmount=on rpool_$INST_UUID/$INST_ID/DATA/default/state
for i in {/etc/nixos,/etc/cryptkey.d}; do
  mkdir -p /mnt/state/$i /mnt/$i
  mount -o bind /mnt/state/$i /mnt/$i
done

zfs create -o mountpoint=/ -o canmount=noauto rpool_$INST_UUID/$INST_ID/ROOT/empty
zfs snapshot rpool_$INST_UUID/$INST_ID/ROOT/empty@start

for i in {home,home/empty}; do
    zfs create -o canmount=on rpool_$INST_UUID/$INST_ID/DATA/default/$i;
done
zfs snapshot rpool_$INST_UUID/$INST_ID/DATA/default/home/empty@start

for i in ${DISK}; do
    mkfs.vfat -n EFI ${i}-part1
    mkdir -p /mnt/boot/efis/${i##*/}-part1
    mount -t vfat ${i}-part1 /mnt/boot/efis/${i##*/}-part1
done
```

And here is my version that causes the install to fail.
```bash
zfs create -o canmount=off -o mountpoint=none bpool_$INST_UUID/$INST_ID
zfs create -o canmount=off -o mountpoint=none bpool_$INST_UUID/$INST_ID/BOOT
zfs create -o mountpoint=/boot -o canmount=noauto bpool_$INST_UUID/$INST_ID/BOOT/default
zfs mount bpool_$INST_UUID/$INST_ID/BOOT/default

zfs create -o canmount=off -o mountpoint=none   rpool_$INST_UUID/$INST_ID/DATA
zfs create -o mountpoint=/ -o canmount=off      rpool_$INST_UUID/$INST_ID/DATA/local
for i in {nix,}; do
    zfs create -o canmount=on -o mountpoint=/$i rpool_$INST_UUID/$INST_ID/DATA/local/$i
done

zfs create -o mountpoint=/ -o canmount=off      rpool_$INST_UUID/$INST_ID/DATA/default

zfs create -o canmount=on                       rpool_$INST_UUID/$INST_ID/DATA/default/state
for i in {/etc/nixos,/etc/cryptkey.d}; do
  mkdir -p /mnt/state/$i /mnt/$i
  mount -o bind /mnt/state/$i /mnt/$i
done

for i in {home,home/empty}; do
    zfs create -o canmount=on rpool_$INST_UUID/$INST_ID/DATA/default/$i;
done
zfs snapshot rpool_$INST_UUID/$INST_ID/DATA/default/home/empty@start

zfs create -o canmount=off -o mountpoint=none rpool_$INST_UUID/$INST_ID/ROOT
zfs create -o mountpoint=/ -o canmount=noauto rpool_$INST_UUID/$INST_ID/ROOT/default
zfs mount rpool_$INST_UUID/$INST_ID/ROOT/default

zfs create -o mountpoint=/ -o canmount=noauto rpool_$INST_UUID/$INST_ID/ROOT/empty
zfs snapshot rpool_$INST_UUID/$INST_ID/ROOT/empty@start

for i in ${DISK}; do
    mkfs.vfat -n EFI ${i}-part1
    mkdir -p /mnt/boot/efis/${i##*/}-part1
    mount -t vfat ${i}-part1 /mnt/boot/efis/${i##*/}-part1
done
```

The resulting failure looks something like this:
```bash
installing the GRUB 2 boot loader on /dev/disk/by-id/nvme-SKHynix_HFS512GDE9X081N_CD13N445112003740...
Installing for i386-pc platform.
/nix/store/xh6bk3yxm6zlyp3vql9jzw39gfvysw43-grub-2.06/sbin/grub-install: warning: cannot open directory `/nix/store/xh6bk3yxm6zlyp3vql9jzw39gfvysw43-grub-2.06/share/locale': No such file or directory.
/nix/store/xh6bk3yxm6zlyp3vql9jzw39gfvysw43-grub-2.06/sbin/grub-install: error: unknown filesystem.
/nix/store/awx3i3k63r99bpk9nygmnk6k0n2dq15c-install-grub.pl: installation of GRUB on /dev/disk/by-id/nvme-SKHynix_HFS512GDE9X081N_CD13N445112003740 failed: No such file or directory
umount: /mnt/boot/efi: not mounted.
cannot unmount '/mnt/boot': unmount failed
cannot unmount '/mnt/nix': unmount failed
```
The last line can change depending on the datasets created but the rest is generally the same. Previously, I had created all the datasets as shown in the documentation and it would fail to unmount `/var/spool`. In the example above, I removed the extra datasets and it failed differently. But I think this is just a red herring and the real reason is somewhere in the order/mounting of pools and datasets.

I rewrote the way the datasets were created and mounted for my own readability. The first chunk creates boot pool datasets, then root pool datasets, etc. But this obviously has undesired side effects. I am uncertain as to the exact reason why it fails, but maybe you have a better idea.  I do notice two things that might be problematic but my knowledge of ZFS is still small.

One possibility I am considering is the time when the root pool is mounted. In your script
```bash
zfs mount rpool_$INST_UUID/$INST_ID/ROOT/default
```
is performed early on and later followed by
```bash
zfs create -o canmount=on rpool_$INST_UUID/$INST_ID/DATA/default/state
for i in {/etc/nixos,/etc/cryptkey.d}; do
  mkdir -p /mnt/state/$i /mnt/$i
  mount -o bind /mnt/state/$i /mnt/$i
done
```
but in mine it happens in reverse.

Another possibility is that a snapshot of the root pool is taken before it is mounted in my script.

Also, for reasons unknown to me, if I use systemd instead of grub and unmount `/mnt` as shown in the PR, I have no issues.

ne9z:
The mounting order is wrong in your script.

You must mount / before /boot.

```sh
zfs mount rpool/.../default
zfs mount bpool/.../default
```

In your script, this order is reversed.

```sh
zfs mount bpool/.../default
zfs mount rpool/.../default
```

Again, I can only guarantee the correctness of the procedures as described in the guide.
If someone does not follow the guide verbatim, then I can't help troubleshooting that.

> systemd

Do you mean systemd-boot? We will never use systemd-boot here.  GRUB is the only bootloader with support for fault-tolerance, meaning with a mirrored setup, you could just remove one disk and expect the system to still boot.

ne9z:
If you mount /boot before /, then I honestly don't know what would happen.  It is a miracle that your system even booted afterwards.

djacu:
Wow that is a major oversight on my end (re: mounting `/boot` before `/`). I was so wrapped up trying to understand ZFS, I forgot the basic fundamentals. Thanks for pointing that out!

Yes, I meant systemd-boot. The [NixOS Wiki](https://nixos.wiki/wiki/ZFS#Mount_.2Fboot) uses systemd-boot but it requires external syncing. I assume, with the way things are setup in the OpenZFS docs, that GRUB does this automatically? Can you elaborate more about the fault-tolerance support that GRUB provides? Is it related to the `grub.extraPrepareConfig` and `grub.extraInstallCommands` options setup in the `zfs.nix` file?

djacu:
Also, what would you like to do regarding this PR? I believe we agree that `umount /mnt/boot/efi` should be removed but the branch name and discussion are overshadowed by extra discussion. Happy to submit a new PR, or modify my original comment, fix the files I modified, and stick to this PR.

ne9z:
First, I'd like to stress the importance of protecting the integrity of contents located under /boot (excluding bootloader program located in /boot/efis and the boot sector).

/boot contains version-controlled initrd, without which one absolutely can not boot the system.  If we do not use ZFS for /boot, there is no protection for its integrity at all.  FYI, the only filesystem systemd-boot supports is vfat (using drivers inherited in UEFI firmware).  This is a non-starter.  

GRUB, in contrast, ships with its own read-only ZFS driver specially designed for booting from ZFS.  This driver fully supports ZFS redundancy: mirror, raidz{1,2,3}.

Another aspect is authentication and encryption for /boot.  Ideally, we should encrypt initrd to ensure that it has not been tempered: on your computer I theoretically can just replace the initrd with a malicious one that sends keyboard inputs (pool password) to me.  Systemd-boot does not support this at all.  But GRUB does.

Also, if one opt to use hibernation with encrypted persistent swap, the key file will be stored alongside initrd.  This necessitates the encryption of /boot. (This feature was removed from the guide due to NixOS bug, see #265).  Again, this rules out systemd-boot.

ne9z:
Also, systemd-boot is EFI only.  GRUB supports not only legacy BIOS, EFI but also embedded bootloaders like u-boot and other architectures like arm, mips, risc, powerpc.

In short, using systemd-boot only gives us limitations and disadvantages.

ne9z:
You might have noticed that, the NixOS Wiki guide didn't even care about bpool: just put /boot on vfat, use easy systemd-boot and be done with it.

ne9z:
As everything else (/ and /boot) is managed by ZFS, we only need to deal with the mirroring of the bootloader: GRUB.

By default, in order to ensure that your installation will boot, the guide specified both BIOS and UEFI versions of GRUB to be installed.

BIOS version of GRUB will be installed to the boot sector of every disk specified in the grub.devices option, so this one is easy.

For EFI, the bootloader is a single grubx64.efi (efi variable writable) file located under /boot/efis/primary-disk-part1/EFI/NixOS/ path.  

If the EFI variable is not writable, (I assume this by default in the guide, there are just too many faulty implementations of EFI in the firmware), it will be installed to the fallback path of /boot/efis/EFI/BOOT/bootx64.efi which every computer will recognize.

We need to copy this bootloader file to every other EFI partition inside /boot/efis/. grub.extraInstallCommands is used for this.

One thing peculiar about EFI partitions is that they use the noauto mount option: they are not mounted at boot, but on access.  Especially with HDD, the paths under /boot/efis/part1 will only become available with a delay.

This delay will cause extraInstallCommands to fail.  Therefore we need grub.extraPrepareConfig to mount these lazy-mounting partitions.

ne9z:
I suggest we use this PR as an opportunity for removing unused commands for /boot/efi (not efis).

The existing line in the PR is one instance.  There is another one on the system-configuration page, inside grub.extraPrepareConfig option.  We should remove that as well.  You can rename the title of this page and reword commit message as needed.

ne9z:
Additional info for hibernation:  with a separate swap partition, hibernation can be used with Root on ZFS systems.  The statement in NixOS wiki unnecessarily scares people away from hibernation altogether.

In general, the various wikis, such as Archwiki and NixOS wiki, contain many outdated or false information.  Use them with caution.

djacu:
> There is another one on the system-configuration page, inside grub.extraPrepareConfig option. We should remove that as well.

There is another instance of /boot/efi being mounted and unmounted in the Encrypted boot pool section of the optional-configuration page. Any advice on whether these need to stay or be removed?

djacu:
The `/boot/efi` in the `mkdir -p /boot/efis /boot/efi` command under the `extraPrepareConfig` should also be removed correct? I've already removed the `mount /boot/efi` line that follows this.

ne9z:
Thanks!  I would never spotted the issue myself since the commands do not have any negative effects on the installation.

I think we should squash the commits before merging.

djacu:
@ne9z thank you for the detailed response about grub and general boot loader details. The PR you submitted to update the docs last year is fantastic and a great help to newcomers like me. However, I've been wondering if some additional details would be helpful to new users. Or alternatively, pointing to some sources that would help people get on a good path for learning.

For example, I was very confused by the dataset layout when I first started reading the docs. Specifically the creation of the `default` and `sys` datasets. The layout was significantly more complicated and involved compared to other guides I had read. I picked up a copy of Lucas and Jude's "FreeBSD Mastery: ZFS" and it explains in Chapter 1 using `default` with a companion dataset makes for safe system upgrades. And the use of `sys` I inferred by reading the other OS installs and assumed that it was for multiboot setups. I remember finding an early draft of your PR on the [NixOS wiki](https://nixos.wiki/index.php?title=User:2r/NixOS_on_ZFS&oldid=5406) and noticed there was some small additional details that didn't make their way into the OpenZFS docs. There are still things I don't understand about grub and your usage of it nor do I have an idea where to start other than reading the GRUB manual.

The knowledge you've dropped in this PR about why GRUB is chosen over systemd-boot is invaluable and I think would be a great addition to the docs. I understand that adding too much detail could likely end up being clutter. But even something small like a footnote pointing to some reference would be amazingly helpful. So with all that said, I'm curious what your thoughts are.

Also happy to open an issue so we can discuss this further without adding any more clutter to this PR if you'd like.

djacu:
> Thanks! I would never spotted the issue myself since the commands do not have any negative effects on the installation.

Sure thing! Happy to help polish these docs as they've helped me so much.

> I think we should squash the commits before merging.

I squashed my 4 commits and updated the commit message.

ne9z:
> I squashed my 4 commits and updated the commit message.

LGTM, @gmelikov This PR is ready for merge.

> details versus clutter

I have to admit that, after a 6-month marathon with Root on ZFS in 2021, providing support for Arch, Fedora, RHEL and NixOS, and attempted to support Alpine and Void Linux, and authoring the boot environment manager bieaz, I got a bit exhausted and did not bother to add the details.  Maybe I should do this eventually.

I also don't know whether anyone would have interest in the details: most would just copy-paste the commands and skip the details, I think.  You can see that my guides are more automated than Debian/Ubuntu guides.

gmelikov (approved):

ne9z:
A side note: I find the concept of a stateless system exceedingly practical and therefore extended this to my home directory `~/` also.  The `/home/empty` dataset in my script was added for this purpose.

- Rollback `~` at reboot.
- `~/.config/` and `~/.emacs.d/` folders are checked into a git repo.  Perform a hard repo reset at every reboot.
- `~/Downloads` and `~/Documents` folders and the `~/.git` repo are bind mounted from `/state/home/User/{Downloads,Documents,.git}`.
- I'm using Sway window manager with greeted as login manager, with the following config

```nix
{
  boot.initrd.postDeviceCommands = ''
    zpool import -Nf rpool_${poolUuid}
    zfs rollback -r rpool_${poolUuid}/nixos/ROOT/empty@start
    zfs rollback -r rpool_${poolUuid}/nixos/DATA/default/home/empty@start
  '';
  services = {
    greetd = {
      enable = true;
      package = pkgs.greetd.greetd;
      vt = 1;
      settings =
        {
          default_session = {
            command = "${pkgs.greetd.greetd}/bin/agreety --cmd 'bash -c /state/home/User/launch.sh'";
          };
        };
    };
  };
}
```
and launch.sh:
```sh
#!/usr/bin/env bash
git -C /home/User reset --hard
home-manager switch
sway
```

This way the empty home will be repopulated from the git repo upon login, by home-manager.

A stateless home has other benefits other than keeping config in track:  browser data will be wiped upon reboot to reduce opportunities of being tracked, for example. [See here for my firefox config for home-manager](https://gist.github.com/ne9z/a01a0787ab18a7f1618f26ec28a983a3).

I also deployed NixOS on my home server to avoid using "cloud" dis-services.
I'm currently studying in Germany while my home is in China, so the home server has to be highly-available without human intervention.  Thus the emphasis on redundancy and auto fail-over in the guide.

As the internet service provider does not provide publicly-accessible IP addresses for private users (NAT), and IPv6 support in China is spotty at best, I have to use some kind of overlay network: Tor, I2P, Yggdrasil, all three of them.  The fact that I can stream 2160p home videos over Yggdrasil speaks volumes for the performance and ingenuity of the developers.  [NixOS guide](https://gist.github.com/ne9z/912773cba81fd638b8541b198f1a8479).

End-to-end overlay network, which grants every computing device equal ability of receiving incoming and outgoing connections, also makes the internet a more just place by demolishing the dichotomy of client-server model, with the latter model led us to today's dystopian landscape of several mega-corporations dominating the internet.

(Maybe I should start a blog someday...)

djacu:
> I also don't know whether anyone would have interest in the details: most would just copy-paste the commands and skip the details, I think.

I know a younger version of myself would have just copied the scripts and hoped everything worked. Currently, and I think there are plenty of people like me, I want to understand how my system works and what the author's rationale was when making certain decisions in the guide. More experienced users will likely understand immediately, but new users might not and having that logic spelled out is a great teaching/learning opportunity. I really like [DVC's](https://dvc.org/doc/start) approach to documentation where all the basic steps are visible, but if you want deeper understanding, you can click on the blue boxes with the gear icon. Not sure if Sphinx supports this but I thought it was neat.

I have to ask because I'm still swimming in a sea of bootloader knowledge, but if you had a recommendation for learning about GRUB and/or bootloaders in general, I would gladly appreciate it. I think I'm set of ZFS as Lucas and Jude's book is concise and extremely informative.

> I find the concept of a stateless system exceedingly practical and therefore extended this to my home directory `~/` also.

I saw that! That's a pretty cool setup. I might copy that once I get more comfortable.
And your home server setup sounds really interesting... more things to learn with time.

> (Maybe I should start a blog someday...)

I'd read it! :)

ne9z:
> I really like DVC's approach to documentation ... click on the blue boxes with the gear icon.

I get the idea.  This particular example didn't work for me -- since I disabled Javascript -- but we could use some CSS for sure.  Will explore options to add details to the guide.  Thanks for the tip!

> GRUB

I have to admit that I'm also not at all familiar with GRUB internals.  It's quite complex: essentially a minimal operating system in fact.  I got familiar with the user-facing aspect of GRUB mostly through experiments:
- While writting the [bieaz](https://gitlab.com/m_zhou/bieaz) boot environment manager for non-Nix distros, I had to read, line by line, the contents of [10_linux](https://git.savannah.gnu.org/cgit/grub.git/tree/util/grub.d/10_linux.in) and https://github.com/Antynea/grub-btrfs to understand how GRUB generates its config and which parameters are needed to boot.
- Later I attempted to add support of booting live ISO directly from GRUB: one can put a pristine live ISO inside EFI partition and tell GRUB to boot from it.  [Only implemented for Arch](https://openzfs.github.io/openzfs-docs/Getting%20Started/Arch%20Linux/Root%20on%20ZFS/4-optional-configuration.html#boot-live-iso-with-grub).  This got me more familiar with how to write GRUB config.

In short, I don't have a recommendation besides doing experiments yourself, which is of course time-consuming.  To save you some time, I can open a issue in your Root on ZFS repo and answer guide-related questions there.

> blog

Actually I have the idea of setting up a non-profit dedicated to promote free/libre software after completing my degree: the website https://chassell.org
For now it's just filled with some of my incoherent ramblings.

PR #297, opened 2022-04-23
Added note about boot pool size consideration; fixes #296

### LINKED ISSUE #296: Root on ZFS - consider large boot partition (Debian)?
At least for the Debian guides, I have found that there are situations where the suggested 1G default for the boot pool may not be sufficient.  While 1G sounds fine and is large enough for doing the basic install, I've found a scenario where it is less than ideal.

In the Debian guides it suggests doing:
```
sgdisk     -n3:0:+1G      -t3:BF01 $DISK
```

So the scenario I found myself in

- I take hourly, daily, and one weekly snapshot of my system zfs partitions.
- I had three different kernels installed on my system (`unattended-upgrades` installed three - I just hadn't rebooted and cleared any old kernels)
- Then a zfs package upgrade was available.

With three kernels, that meant that I had three initrd files of ~85MB each, there were snapshots of those and then the initrd re-generation process takes a backup file to the `/boot` partition so it is essentially storing 3 from when the kernel was installed with the old zfs package in a snapshot, 3 backup initrd files, and then 3 more newly generated initrd files which in itself is using over 750MB.  Long story short, I ran out of disk space, leaving the system in a tricky state where the initrd wasn't re-created so if I had rebooted, I might have been in a bad situation.

I was able to get myself back into a fine state but it just involved a bit more manual user intervention that I wanted.  For now, I just disabled kernel updates via `unattended-upgrades` since I don't automatically reboot anyway so at least I won't be getting new kernel packages and 1G is still fine for a single kernel + snapshots if I need to re-generated my initrd, like in the case of a new zfs package but I'm somewhat now stuck with my current layout without doing some surgery to change it.

I'm not sure if changing the default of `1G` makes sense but would it make sense to add a sizing consideration note to increase it if one intends to automatically snapshot that boot partition?  I know this is something that sounds obvious but I regularly forget about the space that snapshots take up when thinking about sizing and being someone who only had a single partition in my previous, non-zfs on root setup, I wasn't as familiar with the disk space challenges that might come up since I had my whole root partition's space available for `/boot` previously so sizing /boot wasn't much of an issue in the past for me.  I'd be happy to create a PR for that on the Debian guides if that seems like a logical approach.

mbentley:
Signed-off-by: Matt Bentley <mbentley@mbentley.net>

### DIFF
--- a/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
@@ -159,6 +159,11 @@ Step 2: Disk Formatting
      virtio; otherwise, read the `troubleshooting <#troubleshooting>`__
      section.
    - For a mirror or raidz topology, use ``DISK1``, ``DISK2``, etc.
+   - When choosing a boot pool size, consider how you will use the space. A
+     kernel and initrd may consume around 100M. If you have multiple kernels
+     and take snapshots, you may find yourself low on boot pool space,
+     especially if you need to regenerate your initramfs images, which may be
+     around 85M each. Size your boot pool appropriately for your needs.
 
 #. If you are re-using a disk, clear it as necessary:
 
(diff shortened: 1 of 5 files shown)


rlaager:
Thanks!

PR #298, opened 2022-04-26
Added --new-install to tasksel in Debian * Root on ZFS

vacaboja:
@rlaager In the install instructions for Debian with a ZFS root, it might be advisable to add the --new-install switch to tasksel, upon step 8 "full software installation". As far as I understand it, this switch makes tasksel include the "standard system utilities" task, which is supposed to be installed, and would otherwise be skipped.


rlaager:
I ran `tasksel --test --new-install` on an existing server system. It defaulted "Debian desktop environment" to on, and wanted to install task-desktop.  It also defaulted "print server" to on.

So I added a note about that and merged this manually as 5543afab032fcdf3d84d551ec50c7bae2d2eb55c. Thanks!

PR #299, opened 2022-04-27
added mkdir /etc/dropbear-initramfs

JOduMonT:
/etc/dropbear-initramfs is not create automatically with: apt install --yes --no-install-recommends dropbear-initramfs
- at least on Debian Testing ;)


rlaager:
Merged in d8e15645f6ffef758606839d09cc1cf95afbfc29. I used `mkdir -p` just in case the postinst does create it (e.g. in the future).

PR #306, opened 2022-05-18
do *not* enforce UTF-8

anarcat:
i have tried to convert a real, live workstation to ZFS and had to recreate all filesystems because of this silly settings. It actually failed to copy [source code from the supysonic project](https://github.com/spl0k/supysonic/tree/270fa9883b2f2bc98f1482a68f7d9022017af50b/tests/assets/%E6) (which was [eventually removed](https://github.com/spl0k/supysonic/pull/183), but still). I don't think we should suggest such an advanced setting by default and, in general, I find that this guide suggests too many exotic things, instead of focusing on "just install the thing with ZFS".

### DIFF
--- a/docs/Getting Started/Arch Linux/Root on ZFS/2-system-installation.rst
+++ b/docs/Getting Started/Arch Linux/Root on ZFS/2-system-installation.rst
@@ -54,7 +54,6 @@ System Installation
         -O canmount=off \
         -O compression=lz4 \
         -O devices=off \
-        -O normalization=formD \
         -O relatime=on \
         -O xattr=sa \
         -O mountpoint=/boot \
@@ -87,7 +86,6 @@ System Installation
            -O canmount=off \
            -O compression=zstd \
            -O dnodesize=auto \
-           -O normalization=formD \
            -O relatime=on \
            -O xattr=sa \
            -O mountpoint=/ \
@@ -116,6 +114,8 @@ System Installation
      of why requiring UTF-8 filenames may be a bad idea, see `The problems
      with enforced UTF-8 only filenames
      <http://utcc.utoronto.ca/~cks/space/blog/linux/ForcedUTF8Filenames>`__.
+     A previous version of this guide suggested this setting, but it was 
+     reverted because of those problems.
    - ``recordsize`` is unset (leaving it at the default of 128 KiB). If you
      want to tune it (e.g. ``-o recordsize=1M``), see `these
      <https://jrs-s.net/2019/04/03/on-zfs-recordsize/>`__ `various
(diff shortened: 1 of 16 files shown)


rlaager:
> * the `bpool` has a _lot_ of settings, not sure they are all relevant

The features settings on the bpool can be simplified with `compatibility=grub2` once that feature lands in the distros. I've made a note about this.

Though in looking at this, I probably should submit a PR to expand the `grub2` features list.

> * `    -O encryption=aes-256-gcm` is the default, AFAIK

It didn't used to be. That said, `encryption=on` isn't that much simpler. It's still a setting to set. It's arguably useful to use the explicit algorithm name so people can see what it is using and change it if they want. But I can change this to `encryption=on`.

> * `dnodesize=auto` and the xattr stuff are optimisations, maybe they could be left out? or at least regrouped?

I don't know what you mean here. xattrs are enabled by default on modern systems (because systemd wants them), and we should store them using the more efficient option, right? Setting `dnodesize=auto` is required to actually take advantage of the `large_dnode` feature; the default is `legacy`.

> * there are also a _lot_ of mountpoints created, it's a bit excessive

Unfortunately, the FHS intermingles system and user data. The trade-off here is technical correctness vs simplicity.

The list expanded somewhat when Ubuntu added support in the installer and wrote zsys (which is now basically dead). I spent a _lot_ of time discussing this with Canonical. They said they had a lot of experience with this from efforts to make Ubuntu run on mobile devices with read-only root filesystems.

Many of them are optional, if you're not using the thing. Now that I'm moving away from zsys, I should perhaps separate those out (like I used to).

anarcat:
i rerolled this to include all known incantations of `normalization=`, this is now ready for review.

ne9z:
I'm the maintainer of the arch, NixOS, RedHat and Fedora guides (derived partially from the Debian guide) and I can not agree with the premise of this pull request.

[The blog post](http://utcc.utoronto.ca/~cks/space/blog/linux/ForcedUTF8Filenames) explicitly deals with **existing installations** and does not apply to the guides at all, which deal with **new installations** and therefore should be removed.  New installations should enforce UTF-8, see below.

UTF-8 remains desirable, because of interoperability.  The fact that the problematic file path mentioned in the [original description](https://github.com/spl0k/supysonic/pull/183), was eventually removed speaks substance.  In that pull request, the malformed UTF-8 string caused:

- on Debian, scripts to break
- breaks Mac OS (which ZFS also aims to support)

Therefore, I think using UTF-8 is a good default for most users.

If you insist on having malformed paths in your filesystem, you can simply adjust the guide to your own needs, or creating a disk image and experiment on that instead.

We are not **enforcing** the sensible UTF-8 default on anybody -- we only inform potential users of sensible defaults and documenting the most useful options.  It is up to you to decide whether using the defaults or not.

The UTF-8 option is also a particularly thorny one, in that it must be decided at pool creation time and can not be changed later.

rlaager:
I'm going to consider this "wontfix". I understand that utf8only is not without controversy. But it's been the default in the guide for quite some time and it hasn't generated significant complaints. The Ubuntu developers didn't object to this with the installer work and it apparently hasn't generated complaints for them either. With each passing year, non-UTF-8 encoded filenames are less and less likely. I do point out the other side of this, so people can make their own choices.

Issue #307, opened 2022-05-19
reduce the number of options passed to ZFS


anarcat:
This is a rather broad question that I brought up in #306, but it seems to me it deserves at least some discussion.

What I wrote in the PR there is:

> In general, I find that this guide suggests too many exotic things, instead of focusing on "just install the thing with ZFS".

Then I went on with some examples:

 * [ ] the `bpool` has a *lot* of settings, not sure they are all relevant (@rlaager suggested using `compatibility=grub2` once that feature lands, but when is that?)
 * [ ] ~~`-O encryption=aes-256-gcm` is the default, AFAIK~~ (actually fixed in 903d6310335ff0715e7620cca87b95774561430f)
 * [ ] `dnodesize=auto` and the xattr stuff are optimisations, maybe they could be left out? or at least regrouped?
 * [ ] there are also a *lot* of mountpoints created, it's a bit excessive

In general, I think those guides are great and do the right thing. But for an experience sysadmin like me, they are a bit scary, because I feel I need to lookup *everyone* of those single settings to see if they apply to me or not. Since many are not necessarily relevant (like `encryption=X` where X is the default), or are opinionated (like #306 AKA `utf8only`), it can be confusing to figure out which settings should be reviewed and why.

When I first followed the guide, I setup my pools with `utf8only` and now I really regret it, because it's a major interoperability concern. Now #306 should fix this in the future, but I wonder how many more of those creep in there.

In general, for actual necessary cases like `xattr`, shouldn't that be made *default* in OpenZFS upstream, so that we eventually *don't* have to know the magic formula?

In my mind, OpenZFS should converge to having basically *zero* custom arguments. Right now, the root pool creation in the bullseye guide is:

       zpool create \
           -o ashift=12 \
           -O acltype=posixacl -O canmount=off -O compression=lz4 \
           -O dnodesize=auto -O relatime=on \
           -O xattr=sa -O mountpoint=/ -R /mnt \
           rpool ${DISK}-part4

I think it should be turned into this:

       zpool create \
           -O canmount=off -O mountpoint=/ -R /mnt \
           rpool ${DISK}-part4

That means, for example:

- [ ]  the `ashift` detection problem is fixed (!?) -- i know how that sounds, but we can dream, can we?
- [ ]  `posixacl` turns on `dnodesize=auto` and `xattr=sa` automatically, and is automatically enabled on systemd-enabled Linux systems
[...]

PR #308, opened 2022-05-20
WIP: Reorder and regroup options

anarcat:
So it seems most of my ideas of #307 were refused, which is fine. But there's one final idea that I think might reach consensus here, and it's that options should be regrouped together consistently and logically. For example, the `posixacl`, `xattr` and `dnodesize` options go together, `relatime` and `normalization` do not.

This PR attempts to clean that up in one example instructions, the Bullseye one that I am more familiar with. Ideally this would be done elsewhere, but I'm not going to waste any time on *that* until this is somewhat agreed on here. :)

Thanks!

### DIFF
--- a/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
@@ -227,7 +227,8 @@ Step 2: Disk Formatting
 
      zpool create \
          -o cachefile=/etc/zfs/zpool.cache \
-         -o ashift=12 -o autotrim=on -d \
+         -o ashift=12 \
+         -o autotrim=on -d \
          -o feature@async_destroy=enabled \
          -o feature@bookmarks=enabled \
          -o feature@embedded_data=enabled \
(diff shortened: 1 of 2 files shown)


rlaager (changes requested):
This seems fine to me in principle.

- Please make similar changes to the Ubuntu 22.04 HOWTOs. They're close enough that often this can be `git diff Debian/*Bullseye* | patch "Ubuntu/Ubuntu 22.04 Root on ZFS.rst"` If the patch applies cleanly (or you want to do it by hand), you could do Debian Buster and Ubuntu 20.04 too, but that's less important. My main goal is to keep the current versions up-to-date.
- Please add `Signed-off-by` to the commits.
- `dnodesize=auto` is intentionally not used on the bpool. Note that `large_dnode` is not enabled there. (This is slightly harder to tell now that we're using `compatibility=grub2`.)
- I see you reordered in the `Notes` section to keep them in the order they are listed in the command. Do you think we should keep using the order listed, or alphabetize them in the notes? Either way will require moving some options.

rlaager:
Merged as 726f07a. Thanks!

PR #311, opened 2022-05-28
Verify that zpool.cache is aware bpool and rpool [Debian Bullseye Tutorial]

NiDaKuDE:
Hey there,
I had some problems in the past with the Debian Bullseye ZFS on Root tutorial, where I suddenly ended up with a read-only filesystem during "Fix filesystem mount ordering". This extra step fixed the problem for me.

Some documentation told me to mention @rlaager ;-)

### DIFF
--- a/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
@@ -530,6 +530,15 @@ Step 3: System Installation
    An alternative to using ``debootstrap`` is to copy the entirety of a
    working system into the new ZFS root.
 
+#. Verify that zpool.cache is aware of both, bpool and rpool::
+
+      strings /etc/zfs/zpool.cache | grep -P '^rpool|bpool$' | uniq
+
+   In case the command does not output both pools::
+
+      zpool set cachefile=/etc/zfs/zpool.cache rpool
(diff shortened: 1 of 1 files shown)


rlaager:
The read-only filesystem issue was (as far as I know) fixed in commit f110d5248626e341542df2b0b9eb7191fa607305. My co-worker ran into it, and we eventually tracked this down. As the commit message says, I have no idea why I added those commands in the first place. So I just removed them.

The cachefile situation is a bit of a mess here. I'm explicitly asking for it to be set when the bpool is created, but only on the bpool. But then I have the user create a service that imports it with `cachefile=none`.

NiDaKuDE:
The last time I followed the instructions and forgot to run the commands I added, was at least two weeks ago. So it is likely that it was fixed by your mentioned issue.I ran into the problem a while ago (shortly after the first Bullseye-Version of the Documentation was released) and I was like "Someone surely has reported/fixed this issue already.". Like a month ago, I wrote myself a Post-It to remember opening a PR. In the end I forgot to check the last commits before opening a PR. Sorry about that. ;-) 

Another "interesting" observation I made is that the order in which I run the two `zpool set cachefile=` commands does matter. If I run the command for the bpool prior to the one for the rpool, only one of both ends up in the cachefile. If I change the order,  (rpool then bpool) both end up in the cachefile. One nice touch is that the "zed" portion works first try ;-)

PR #312, opened 2022-05-30
Update RHEL 8-based distro Root on ZFS.rst

furicle:
Matches terminology used elsewhere, instead of Cent/Rocky/Alma/Oracle etc etc


ne9z:
Thanks for the contribution!

However, the explicit reference to Rocky Linux is to make it clear on
the compatibility and limitation of the guide -- it is only tested with
Rocky Linux.

To say otherwise would be tantamount to lying, in my opinion.


It would be great if we have the time to test the guide on other
distros, and only then, expand the scope to the entire family of RHEL
derivatives.

PR #319, opened 2022-06-14
Updates Root on ZFS guide 

Asheboy:
I've removed the part at the beginning of the guide which stated that ZFS was almost removed from the 22.04 installer. I'm pretty sure this wasn't the case and the belief of this came about because of the incorrectly named [launchpad issue](https://bugs.launchpad.net/ubuntu/+source/ubiquity/+bug/1966773). This mistakenly referenced ZFS instead of zsys. See the below update to the title of the issue:

image

### DIFF
--- a/docs/Getting Started/Ubuntu/Ubuntu 22.04 Root on ZFS.rst
+++ b/docs/Getting Started/Ubuntu/Ubuntu 22.04 Root on ZFS.rst
@@ -12,9 +12,7 @@ Overview
 Ubuntu Installer
 ~~~~~~~~~~~~~~~~
 
-The Ubuntu installer still has ZFS support, but `it was almost removed for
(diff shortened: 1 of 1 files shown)


rlaager:
seb128's later comment notwithstanding, if you read the actual commit that was proposed on that bug, it _was_ about removing all ZFS support from the _installer_. Note that this is not the same as removing support for _ZFS_. In other words, there are three levels of possible removal:

1. Remove zsys. Keep ZFS installer support. Keep ZFS.
2. Remove zsys and ZFS installer support. Keep ZFS.
3. Remove ZFS.

The actual commit was 1. What they ended up doing (supposedly, I haven't actually tested the Ubuntu 22.04 installer) is 2.

I suspect they will end up doing 1 in a future release.

The comment in the guide is correct.

PR #321, opened 2022-06-21
Update 2-system-configuration.rst

mariaa144:
Add correct path to `aliases` file on immutable step


mariaa144:
Is this change to the `touch` command necessary? This is the first time going through the guide and the command failed for me until I modified it to `/mnt/state/etc/aliases`.

ne9z:
Yes, it is.  Thanks for the fix!  @gmelikov

gmelikov (approved):
Thank you!

PR #322, opened 2022-06-26
Update Hardware.rst

gyto6:
Add NVMe et SATA drives with PLP + Optane Drives

### DIFF
--- a/docs/Performance and Tuning/Hardware.rst
+++ b/docs/Performance and Tuning/Hardware.rst
@@ -504,6 +504,7 @@ follows:
 
 -  Intel 750
 -  Intel DC P3500/P3600/P3608/P3700
+-  Micron 7300/7400/7450 PRO/MAX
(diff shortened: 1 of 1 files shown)


docs/Performance and Tuning/Hardware.rst
```diff
@@ -588,6 +589,7 @@ follows:
 -  Intel 710
 -  Intel 730
 -  Intel DC S3500/S3510/S3610/S3700/S3710
+-  Kingston DC450R/DC500R/DC500M
```
gmelikov, line 592: Could you please point at some source? https://www.kingston.com/en/ssd/dc450-data-center-solid-state-drive says at least DC450R doesn't have PLP.

gyto6: Oups, I double checked the DC450R and deleted it but I pulled the wrong commit. My mistake.

gmelikov (approved):
Thank you!

PR #338, opened 2022-08-23
Add cryptsetup-initramfs package to include cryptsetup into initrd.

KaliszAd:
The initrd cannot otherwise decrypt the partition.


gmelikov (approved):
Thank you!

PR #340, opened 2022-08-26
Workload tuning: prepare for FS_CASEFOLD_FL patch

gmelikov:


gmelikov:
@ryao please review

gmelikov:
[#13790](https://github.com/openzfs/zfs/pull/13790) was closed, so close this PR too at least for now.

ryao:
I just saw this now. Actually, there is a small performance improvement from using casesensitivity=insensitive even without that patch whenever it causes a directory lookup hit that would have been a miss. Samba can also benefit from this if you tell it that ZFS is using case sensitivity.

The only issue is that we do not have the exact same definition of case insensitivity as Windows, although we are close.

gmelikov:
@ryao thank you for reinstating my thought, I even think that my wording is right even for unmerged optimizations, merged as is.

ryao:
@gmelikov It might not have been appropriate to continue referencing FS_CASEFOLD_FL. I would have worded this differently, such as saying that it avoids expensive directory misses that could have been hits. There is also the possibility of tuning Samba to assume that ZFS is case insensitive when this is set. It is an optimization that we are not doing with sharesmb.

gmelikov:
@ryao
> It might not have been appropriate to continue referencing FS_CASEFOLD_FL.

FS_CASEFOLD_FL is used here only as anchor name, so it looks ok for me as anchor to FS_CASEFOLD_FL PR.

> I would have worded this differently, such as saying that it avoids expensive directory misses that could have been hits. 

I'll try to get back to `Workload Tuning` page later this month and maybe expand it further, but of course feel free to push PRs with exact wording, github allows to edit rst files via web to make it easier.

Issue #345, opened 2022-09-10
property 'recordsize' is not a valid pool property


bajjer:
@rlaager, I have the following issue with the Ubuntu 22.04 Root on ZFS HOWTO:

I'm attempting to set recordsize=1M as indicated in the HOWTO:

> recordsize is unset (leaving it at the default of 128 KiB). If you want to tune it (e.g. -o recordsize=1M), see [these](https://jrs-s.net/2019/04/03/on-zfs-recordsize/) [various](http://blog.programster.org/zfs-record-size) [blog](https://utcc.utoronto.ca/~cks/space/blog/solaris/ZFSFileRecordsizeGrowth) [posts](https://utcc.utoronto.ca/~cks/space/blog/solaris/ZFSRecordsizeAndCompression).

Command:
`root@ubuntu:~# zpool create \
    -o ashift=12 -o recordsize=1M \
    -o autotrim=on \
    -O encryption=on -O keylocation=prompt -O keyformat=passphrase \
    -O acltype=posixacl -O xattr=sa -O dnodesize=auto \
    -O compression=zstd \
    -O normalization=formD \
    -O relatime=on \
    -O canmount=off -O mountpoint=/ -R /mnt \
    rpool mirror \
    $hdd1 \
    $hdd2 \
    mirror \
    $hdd3 \
    $hdd4 \
    log mirror \
    $nvme1-part2 \
    $nvme1-part2 \
    special mirror \
    $nvme1-part3 \
    $nvme2-part3`

Error:
`property 'recordsize' is not a valid pool property`

ZFS Version:
`version:        2.1.4-0ubuntu0.1`

rlaager:
`recordsize` is a _filesystem_ property. You set filesystem properties with `zfs create -o`, or on the root dataset `zpool create -O` (note capital O).

PR #346, opened 2022-09-10
Revise example to modify recordsize to apply to the root dataset during the rpool creation step

bajjer:
Change the provided example to tune the ``recordsize`` to apply to the root dataset given it's in the notes under the step creating the rpool. See https://github.com/openzfs/openzfs-docs/issues/345#issuecomment-1242657449


rlaager:
Thanks! I applied the same change to all the distro HOWTOs.

bajjer:
Thank you, I see it updated now.

PR #352, opened 2022-09-23
Add double quotes to DISK=$DISK

ne9z:
Without the double quotes, the $DISK variable expands and will not handle multi-disk correctly.

Co-authored-by: @baberlevi


ne9z:
@gmelikov

gmelikov (approved):

PR #363, opened 2022-11-02
FreeBSD Getting Started updates

brooksdavis:
These commits update the FreeBSD instructions and address issues I ran into getting builds from the openzfs repo working.


gmelikov (approved):
Thank you!

PR #366, opened 2022-12-03
Update openSUSE Tumbleweed Root on ZFS.rst

jze2:
When mail dataset is created in /var/mail, the filesystem package will fail to install (as it does not expect /var/mail to be a directory, see https://archive.virtualmin.com/node/23096), so create it in /var/spool/mail instead as is usual.


gmelikov (approved):
Thank you!

PR #369, opened 2022-12-19
improved PostgreSQL advice

shribe:
The advice to not cache data in ARC is simply wrong--PG does not work that way, shared buffers are more special purpose. PG *expects* the majority of caching to be done by the OS file cache and is designed around that assumption.

LZ4 compression changes things with regard to record size--even with fast NVMe--using a record size that will fit multiple compressed blocks increases performance by reducing the total data written despite the partial record writes. (I have benchmarked this extensively.)

### DIFF
--- a/docs/Performance and Tuning/Workload Tuning.rst
+++ b/docs/Performance and Tuning/Workload Tuning.rst
@@ -50,7 +50,10 @@ zfs properties respectively, which can be set on both zvols and
 datasets. Possible settings are ``all``, ``none`` and ``metadata``. It
 is possible to improve performance when a zvol or dataset hosts an
 application that does its own caching by caching only metadata. One
-example is PostgreSQL. Another would be a virtual machine using ZFS.
+example would be a virtual machine using ZFS. Another would be a 
+database system which manages its own cache (Oracle for instance).
+PostgreSQL, by contrast, depends on the OS-level file cache for the 
+majority of cache.
 
 .. _alignment_shift_ashift:
 
@@ -614,9 +617,13 @@ settings must be disabled to disable AIO.
 PostgreSQL
 ~~~~~~~~~~
 
-Make separate datasets for PostgreSQL's data and WAL. Set ``recordsize=8K``
-on both to avoid expensive partial record writes. Set ``logbias=throughput``
-on PostgreSQL's data to avoid writing twice.
+Make separate datasets for PostgreSQL's data and WAL. Set 
+``compression=lz4`` and ``recordsize=32K`` (64K also work well, as 
+does the 128K default) on both. Configure ``full_page_writes = off`` 
+for PostgreSQL, as ZFS will never commit a partial write. For a database 
+with large updates, experiment with ``logbias=throughput`` on 
+PostgreSQL's data to avoid writing twice, but be aware that with this 
+setting smaller updates can cause severe fragmentation.
 
 SQLite
 ~~~~~~


docs/Performance and Tuning/Workload Tuning.rst
```diff
@@ -614,9 +617,13 @@ settings must be disabled to disable AIO.
 PostgreSQL
 ~~~~~~~~~~
 
-Make separate datasets for PostgreSQL's data and WAL. Set ``recordsize=8K``
-on both to avoid expensive partial record writes. Set ``logbias=throughput``
-on PostgreSQL's data to avoid writing twice.
+Make separate datasets for PostgreSQL's data and WAL. Set 
+``compression=lz4`` and ``recordsize=32K`` (64K also work well, as
```
gmelikov, line 621: Do you have any source with benchmarks?

Larger recordsize truly can help with read-mainly workload, but on writes it adds heavily for write amplification. Default recommendations should be well-suited for all cases, even if some of them won't be optimal. Usually default can be changes until one of workload won't degrade drastically. Write amplification is such degradation.

docs/Performance and Tuning/Workload Tuning.rst
```diff
@@ -50,7 +50,10 @@ zfs properties respectively, which can be set on both zvols and
 datasets. Possible settings are ``all``, ``none`` and ``metadata``. It
 is possible to improve performance when a zvol or dataset hosts an
 application that does its own caching by caching only metadata. One
-example is PostgreSQL. Another would be a virtual machine using ZFS.
+example would be a virtual machine using ZFS. Another would be a 
+database system which manages its own cache (Oracle for instance).
+PostgreSQL, by contrast, depends on the OS-level file cache for the
```
gmelikov, line 55: It would be great to have a source here too, or benchmark.

gmelikov (commented):
Thank you for contribution! I'm not a Postgresql guru, but here are some comments.

docs/Performance and Tuning/Workload Tuning.rst
shribe: But it does NOT actually increase write amplification significantly, because of the compression reducing the total amount of data written. Large number of runs with pgbench with many different loads, plus our own internal loads, show significant improvement in both read and write with 16K and 32K blocks, more minor improvements at 64K and 128K, plateau at 26K and 512K, tiny (in the noise) degradation at 1M. Note that I separately timed the initial data load vs the mixed transaction pseudo-TPC portion, and the pure write phase was significantly faster.

Consider 3 cases:

- pure/heavy read load: as you pointed out, of course it helps
- pure/heavy write load: PG is writing multiple blocks at a time, so the write amplification does not come into play
- mixed load: reads are faster, single block updates are slower, multi-block updates are faster

There are a number of blog posts on ZFS tuning for PostgreSQL, and every one that has been written since LZ4 was available comes to the same conclusion.

The advice to use 8K record size was correct prior to LZ4 and provided a significant performance improvement. It is of course still better than 128K record size without compression. But LZ4 + record size larger than PG block size offers even better performance.

docs/Performance and Tuning/Workload Tuning.rst
shribe: Nearly any source regarding PostgreSQL tuning. Start with the documentation, which makes a brief reference to it: https://www.postgresql.org/docs/15/runtime-config-resource.html. 

Also:

https://distributedsystemsauthority.com/optimizing-postgresql-shared-buffers/
https://postgreshelp.com/postgresql_shared_buffers/
https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server
https://redfin.engineering/how-to-boost-postgresql-cache-performance-8db383dc2d8f
https://www.postgresql.org/message-id/20020515112447.C23717%40mail.libertyrms.com
https://www.postgresql.org/message-id/15338.1006185597@sss.pgh.pa.us
https://www.crunchydata.com/blog/optimize-postgresql-server-performance
https://severalnines.com/blog/architecture-and-tuning-memory-postgresql-databases/
https://bun.uptrace.dev/postgres/performance-tuning.html#shared-buffers
https://momjian.us/main/writings/pgsql/hw_performance/
https://gitlab.com/gitlab-org/gitlab/-/issues/258640 (see comments about eviction algo)

ryao:
> The advice to not cache data in ARC is simply wrong--PG does not work that way, shared buffers are more special purpose. PG _expects_ the majority of caching to be done by the OS file cache and is designed around that assumption.

In hindsight, the section on ARC was not well written, which is my fault.

At the time, there was no central write-up on performance tips and I had been helping people do tuning in IRC often enough that I had felt it was beneficial to write it down so that I would not be repeating myself so much. When I wrote it down, I sought to record *everything* I knew, but my knowledge in the area of ARC was somewhat immature, which lead to it being written poorly.

That said, double caching in general, is less than ideal, since the data being duplicated would be better omitted from the further away cache so that things that actually benefit would be included in it. ARC is relatively self tuning, but if you were to use the omission from ARC to increase a VM's memory so that it could include its entire working set in its own LRU cache, that would make any caching by ARC pointless for anything other than VM reboots. That should have been explained, but it was not.

The mention of PostgreSQL there had been more intended to be a possibility than a hard rule, but that was not clearly explained. Also, at the time:

* Far more experienced people had previously informed me that databases used O_DIRECT because they did their own caches and explicitly wanted to avoid having the filesystem do caching because it was pointless.
* PostgreSQL supposedly had its own caching algorithm that was similar to ARC, and superior to what the OS typically did from a mix of what I already knew about the page cache and what I had read in write-ups on it that the PostgreSQL developers had published.
* I had previously used PostgreSQL in a project and had some minor familiarity with it.

It seemed obvious that PostgreSQL was in a similar situation, so a mention seemed merited, but unlike VMs, I had not tested the idea on PostgreSQL, which was a mistake. Of course, the section was so poorly written that even if I had confirmed it, it would still have been problematic in hindsight.

That said, I did omit ARC tuning from the section on PostgreSQL specific advice. That had been consciously done since it was not a hard rule that disabling ARC's data caching on PostgreSQL was beneficial. Even if it was, it might have only benefited the WAL for example.

Finally, I am really happy to see someone volunteer to give this some attention. The documentation is in need of it and it is hard to tackle it alongside all of the other things that I am doing since returning to the project. I have enough things on my todo list it is already difficult for me to get to the things on it in a timely manner, so as much as I would like to add this to my todo list, it just is not realistically possible for me to do that in the foreseeable future.

docs/Performance and Tuning/Workload Tuning.rst
ryao: LGTM

docs/Performance and Tuning/Workload Tuning.rst
ryao: LGTM.

ryao (approved):

shribe:
> That said, double caching in general, is less than ideal, since the data being duplicated would be better omitted from the further away cache so that things that actually benefit would be included in it. ARC is relatively self tuning, but if you were to use the omission from ARC to increase a VM's memory so that it could include its entire working set in its own LRU cache, that would make any caching by ARC pointless for anything other than VM reboots. That should have been explained, but it was not.

Absolutely true, and I'm willing to take this explanation and edit it into this PR.

> Far more experienced people had previously informed me that databases used O_DIRECT because they did their own caches and explicitly wanted to avoid having the filesystem do caching because it was pointless.

That is true of Oracle for sure (well, it's configurable, but generally how it's done for serious databases); I wonder if we could get someone to chime in with knowledge of other DBs?

> PostgreSQL supposedly had its own caching algorithm that was similar to ARC, and superior to what the OS typically did from a mix of what I already knew about the page cache and what I had read in write-ups on it that the PostgreSQL developers had published.

PG's cache is rather special-purpose, helps most with data being actively written, and is deliberately simple and depends on OS cache for the "big" cache. Perhaps I should expand my comment? As well, there is a special case: if the entire working set of the db will fit in RAM, then set shared_buffers big enough to hold it, OS cache no longer needed, and it would probably make sense to set ARC to metadata only. I've never tested that, as "database that will fit in RAM" is not my concern ;-) That might be too much detail, along with how the ARC algo may be better than PG's for sequential scans of tables that don't quite fit in memory (untested theory...)

Also, does my explanatory comment "The advice to use 8K record size was correct prior to LZ4..." actually belong in the doc itself? Same question for anything else in the comments here, mine or yours?

What's the line between "sound default advice with enough explanation for people to have some idea why" vs "out of scope details"?

ryao:
> Absolutely true, and I'm willing to take this explanation and edit it into this PR.

To be clear, I meant that I had a made a mistake when I originally wrote this, not that you had made a mistake in not fixing it. My mistakes are not your mistakes and I did not intend to suggest that they were. Your response made me realize that what I wrote could have been interpreted to mean that, which had not been my intention.

That said, seeing a revision that rectifies my original mistake would be nice. It need not be in this PR and it instead could go into a follow-up PR, but either way is fine with me.

Also, I am inclined to just merge this as is, but @gmelikov seems to have volunteered to be the one to merge this given that he was looking for reviewers. He might have had another reviewer in mind, in which case, so it would be premature to merge it ahead of that other hypothetical other person looking at it. That is the only reason I have not merged it already.

> That is true of Oracle for sure (well, it's configurable, but generally how it's done for serious databases); I wonder if we could get someone to chime in with knowledge of other DBs?

Getting volunteers is somewhat time consuming, since you need to hunt them down to ask them to make time. If you are willing to volunteer to do that, that would be great.

> PG's cache is rather special-purpose, helps most with data being actively written, and is deliberately simple and depends on OS cache for the "big" cache. Perhaps I should expand my comment?

Feel free.

> As well, there is a special case: if the entire working set of the db will fit in RAM, then set shared_buffers big enough to hold it, OS cache no longer needed, and it would probably make sense to set ARC to metadata only. I've never tested that, as "database that will fit in RAM" is not my concern ;-) That might be too much detail, along with how the ARC algo may be better than PG's for sequential scans of tables that don't quite fit in memory (untested theory...)

In-memory databases are a fairly popular concept in recent years, although I am ignorant of their details, since I have not had much reason to look at them.

If you were willing to write something about them, I would be supportive of including that (provided that review does not expose any technical issues in it).

> Also, does my explanatory comment "The advice to use 8K record size was correct prior to LZ4..." actually belong in the doc itself? Same question for anything else in the comments here, mine or yours?

It might. These days, you can often get better performance from by using compression (especially zstd) to increase the amount that can be stored in ARC (although in rare cases, the latency penalty of compressed ARC is not worth it). Sometimes the improvement made by compression of larger records to cache hit rates is enough that even accepting the read-modify-write penalty from having the ZFS record size exceed the database record size is worth it. There could be cases where larger filesystem record sizes for better compression are only worthwhile with larger database record sizes. It is hard to provide a one size fits all answer that considers these possibilities without doing plenty of testing.

Talking about this probably requires doing some background research in the form of testing.

> What's the line between "sound default advice with enough explanation for people to have some idea why" vs "out of scope details"?

That is subjective. All I can say here is do your best and hopefully reviewers will also think about this.

docs/Performance and Tuning/Workload Tuning.rst
gmelikov: > But it does NOT actually increase write amplification significantly, because of the compression reducing the total amount of data written.

The key part here is amplification size, I agree it may affect less than gains, but it will raise

> pure/heavy write load: PG is writing multiple blocks at a time, so the write amplification does not come into play

Could you describe more here? It looks so for WAL, but not for data files IIUC, in case with `UPDATE`s, no?

gmelikov:
Thank you @shribe @ryao !

My goal is to change performance-specific things with some benchmarks (ideally - reproducible) as proofs for PR (but it's only recommended, not truly mandatory). It's great that we had @ryao to get some history context here :)

I'll merge this one as is. Feel free to comment/open issue/make PR with additions.

docs/Performance and Tuning/Workload Tuning.rst
gmelikov: @shribe If you have some benchmarks' data next time - it would be great to have it in PR for history, thank you.

shribe: > Could you describe more here? It looks so for WAL, but not for data files IIUC, in case with UPDATEs, no?

100% true for WAL, less so for tables. But WAL is about 1/2 the writes, so it's important. IIRC, I did not use separate WAL and data in my tests. Hmmm.

PG has slightly more bias toward sequential appends to end of table file than other DBs, because of the MVCC model limiting the conditions under which update-in-place can happen. But reuse of space for known "dead" tuples adds back a degree of random access.

gmelikov: > But reuse of space for known "dead" tuples adds back a degree of random access.

Yep, that's most interesting part of this question for me, because usually OLTP may not grow so fast, but will change (delete/insert or just update), and eventually it will reuse space after VACUUMing, for example (IIRC).

ryao:
@shribe  From Slack:

```
Jason King
  [11:41 AM](https://openzfs.slack.com/archives/C052RGXL5/p1671554497648089)
You may want to add wal_init_zero = false as well for Postgres
[11:42](https://openzfs.slack.com/archives/C052RGXL5/p1671554546027129)
having it true is no benefit on zfs at best, and could potentially be detrimental (needless consuming space)
[11:45](https://openzfs.slack.com/archives/C052RGXL5/p1671554703281689)
(the original premise behind the true behavior was somewhat flawed regardless)
New


Jason King
  [1:01 PM](https://openzfs.slack.com/archives/C052RGXL5/p1671559260314139)
('if we preallocate relying on implementation defined behavior of the fs, we'll never get an error when writing out later')
```

PR #373, opened 2023-01-02
Use nix builtins for `hostId`

concatime:
I think this is more nix-friendly approach, and does not need to spawn a new process.


ne9z:
Yep, that's much better than the current method used.  It also makes the
resultig configuration file (disk configuration excepted) portable
between computers.

gmelikov (approved):

PR #377, opened 2023-01-21
Updated list of groups

christophhagemann:
User permissions for audio etc. are handled by systemd nowadays, adding user to these groups is no longer recommended and will break user-switching in case of audio.

### DIFF
--- a/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
@@ -957,7 +957,7 @@ Step 6: First Boot
 
      cp -a /etc/skel/. /home/$username
      chown -R $username:$username /home/$username
-     usermod -a -G audio,cdrom,dip,floppy,netdev,plugdev,sudo,video $username
+     usermod -a -G sudo $username
 
 #. Mirror GRUB
 


gmelikov:
Could you please point at some source with such recommendations?

For example https://wiki.debian.org/Sound still advices to add `audio` group as troubleshooting.

christophhagemann:
Thanks to the pointer to the debian wiki, I'll file a bug there as well. Took me a lot of time to figure out the problem. (If a user is in the audio group, other user can see no sound device until the audio-group user logs out.) Documented here:
https://wiki.ubuntu.com/Audio/TheAudioGroup

The system doesn't add new users to these groups as well:

$ adduser testuser
$ groups testuser
testuser : testuser users

Audio, plugdevices, cdrom etc. work.

Best regards,
Christoph

BTW: I just set up my machine with native zfs per-user encrypted home with automatic unlock on login and lock on logout (via systemd and PAM). If you'd like to include this variant in your guide, I'd take the time to write up my steps.

rlaager:
> BTW: I just set up my machine with native zfs per-user encrypted home with automatic unlock on login and lock on logout (via systemd and PAM). If you'd like to include this variant in your guide, I'd take the time to write up my steps.

I am definitely interested in that.

rlaager:
I just installed Debian 11 in a VM using the graphical installer. It put the user in the following groups: $username cdrom floppy audio dip video plugdev netdev bluetooth scanner

I was able to play audio. I added another user. I put that user in the audio group. I did "Switch User" in the GUI to switch to that second user. I was able to play audio.

I logged out of the second user. I removed that user from the audio group. I switched back to that user in the GUI. I was able to play audio.

So it seems that being in the audio group is not necessary, but I cannot reproduce that it is harmful, and the installer is still putting the initial user into it.

christophhagemann:
I did some additional testing yesterday and couldn't reproduce my
initial findings. So you're right: it does no harm being in the audio
group. I still wonder if there's a benefit or not being in all those
groups.


Am Samstag, dem 28.01.2023 um 13:54 -0800 schrieb Richard Laager:

christophhagemann:
Am Samstag, dem 28.01.2023 um 13:24 -0800 schrieb Richard Laager:
> > BTW: I just set up my machine with native zfs per-user encrypted
> > home with automatic unlock on login and lock on logout (via systemd
> > and PAM). If you'd like to include this variant in your guide, I'd
> > take the time to write up my steps.
> I am definitely interested in that.
Did a pull request here:
https://github.com/openzfs/openzfs-docs/pull/379

PR #379, opened 2023-01-30
add encryption option: per-user encrypted homdir

christophhagemann:
I am no developer, just a somewhat experienced user. I tried my very best to not do something stupid, but I don't know if I succeeded. It works here (TM). Feel free to change anything.

Please note: I'm on debian/testing, not on bullseye. *Should* make no difference...

### DIFF
--- a/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
@@ -951,10 +951,28 @@ Step 6: First Boot
    Replace ``YOUR_USERNAME`` with your desired username::
 
      username=YOUR_USERNAME
+     
+     Choose one of the following:
+     
+     - Unencrypted homedir or whole-disk encryption
+     
+       ::
 
-     zfs create rpool/home/$username
-     adduser $username
-
+         zfs create rpool/home/$username
+         adduser $username
+       
+     - Encrypted homdir per user, automatic decryption on login
+     
+       ::
+       
+         zfs create rpool/home/$username -o encryption=on -o keyformat=passphrase -o keylocation=prompt -o canmount=noauto
+         zfs mount rpool/home/$username
+         adduser $username
+         
+       **Note**: Use the same strong password for ZFS encryption and user password. Please note: After a breach of your password changing the ZFS password does not restore protection.
+     
+   In either case, continue::
+     
      cp -a /etc/skel/. /home/$username
      chown -R $username:$username /home/$username
      usermod -a -G audio,cdrom,dip,floppy,netdev,plugdev,sudo,video $username
--- a/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
@@ -71,9 +71,9 @@ Contributing
 Encryption
 ~~~~~~~~~~
 
-This guide supports three different encryption options: unencrypted, ZFS
-native encryption, and LUKS. With any option, all ZFS features are fully
-available.
+This guide supports four different encryption options: unencrypted, ZFS
+native encryption, LUKS and per-user encrypted homedirs. With any option,
+all ZFS features are fully available.
 
 Unencrypted does not encrypt anything, of course. With no encryption
 happening, this option naturally has the best performance.
@@ -303,7 +303,7 @@ Step 2: Disk Formatting
 
    Choose one of the following options:
 
-   - Unencrypted::
+   - Unencrypted (choose this if you want per-user encyprtion later on)::
 
        zpool create \
            -o ashift=12 \
--- a/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
@@ -93,6 +93,11 @@ entered at the console. Performance is good, but LUKS sits underneath ZFS, so
 if multiple disks (mirror or raidz topologies) are used, the data has to be
 encrypted once per disk.
 
+Per-user encrypted homdirs encrypts only the datasets for each user. The system
+boots without the need to enter a passphrase. No user-data is accessible until
+the user logs in. The homdirs-datasets are automatically unlocked on login and
+locked on last logout. Mixing encrypted and non-encrypted homdirs is supported.
+
 Step 1: Prepare The Install Environment
 ---------------------------------------
 
--- a/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
@@ -1008,6 +1008,77 @@ Step 6: First Boot
 
          # still here? unlock now
          zfs load-key "rpool/home/$PAM_USER" < /dev/stdin
+         
+       Now create a systemd service to mount our unlocked dataset::
+       
+         mkdir -p /etc/systemd/system/user@.service.d/
+         printf "[Unit]\n\
+                 Requires=user-zfs-mount@%%i.service\n"\
+           > local-mount-zfs.conf
+           
+       Write the service definition::
+       
+         vi /etc/systemd/system/user-zfs-mount@.service
+         
+       with this content::
+       
+         # local service to mount encrypted homdirs
+
+         [Unit]
+         Description=User ZFS mount /home/ for UID %i
+         After=dbus.service
+         StopWhenUnneeded=yes
+         IgnoreOnIsolate=yes
+
+         [Service]
+         ExecStart=/usr/local/sbin/mount-zfs-homedir start %i
+         ExecStop=/usr/local/sbin/mount-zfs-homedir stop %i
+         Type=oneshot
+         RemainAfterExit=yes
+         Slice=user-%i.slice
+         
+       Finally, create the helper script::
+       
+         touch /usr/local/sbin/mount-zfs-homedir
+         chmod a+x /usr/local/sbin/mount-zfs-homedir
+         vi /usr/local/sbin/mount-zfs-homedir
+         
+       And put into it::
+       
+         #!/bin/bash
+
+         set -e
+
+         # called from systemd via /etc/systemd/system/user-zfs-mount@.service
+         # to mount/unmount
+         # we get: $1 - start/stop, $2 - UID
+
+         # get username from UID passed to us by systemd
+         USERNAME=$(id -nu $2)
+
+         # gracefully exit if no such dataset exists
+         zfs list rpool/home/$USERNAME || exit 0
+
+         case $1 in
+         start)
+           # exit if already mounted
+           findmnt "/home/$USERNAME" && exit 0
+
+           # Mount homedir of user we are logging in as
+           zfs mount "rpool/home/$USERNAME"
+         ;;
+
+         stop)
+           # if the dataset of the user logging out is not encrypted, leave it alone
+           [ `zfs list rpool/home/$PAM_USER -o encryption -H` == off ] && exit 0
+         
+           zfs umount "rpool/home/$USERNAME"
+           zfs unload-key "rpool/home/$USERNAME"
+         ;;
+
+         esac
+
+
 
 
      
--- a/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
@@ -952,16 +952,16 @@ Step 6: First Boot
 
      username=YOUR_USERNAME
      
-     Choose one of the following:
+   Choose one of the following:
      
-     - Unencrypted homedir or whole-disk encryption
+   - Unencrypted homedir or whole-disk encryption
      
        ::
 
          zfs create rpool/home/$username
          adduser $username
        
-     - Encrypted homdir per user, automatic decryption on login
+   - Encrypted homdir per user, automatic decryption on login
      
        ::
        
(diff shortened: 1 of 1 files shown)


docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -71,9 +71,9 @@ Contributing
 Encryption
 ~~~~~~~~~~
 
-This guide supports three different encryption options: unencrypted, ZFS
-native encryption, and LUKS. With any option, all ZFS features are fully
-available.
+This guide supports four different encryption options: unencrypted, ZFS
+native encryption, LUKS and per-user encrypted homedirs. With any option,
```
rlaager, line 75: nits: Add Oxford comma; expand "homedirs" to full "home directories".

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -93,6 +93,11 @@ entered at the console. Performance is good, but LUKS sits underneath ZFS, so
 if multiple disks (mirror or raidz topologies) are used, the data has to be
 encrypted once per disk.
 
+Per-user encrypted homdirs encrypts only the datasets for each user. The system
```
rlaager, line 96: typo in "homdirs", but also expand it.
"Per-user encrypted home directories encrypts only the home directory datasets for each user."

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -93,6 +93,11 @@ entered at the console. Performance is good, but LUKS sits underneath ZFS, so
...
+boots without the need to enter a passphrase. No user-data is accessible until
```
rlaager, line 97: nit: drop the dash in user-data

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -93,6 +93,11 @@ entered at the console. Performance is good, but LUKS sits underneath ZFS, so
...
+the user logs in. The homdirs-datasets are automatically unlocked on login and
```
rlaager, line 98: nit: "The home directory datasets"

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -93,6 +93,11 @@ entered at the console. Performance is good, but LUKS sits underneath ZFS, so
...
+locked on last logout. Mixing encrypted and non-encrypted homdirs is supported.
```
rlaager, line 99: nit: s/homdirs/home directories/

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -303,7 +308,7 @@ Step 2: Disk Formatting
 
    Choose one of the following options:
 
-   - Unencrypted::
+   - Unencrypted (choose this if you want per-user encyprtion later on)::
```
rlaager, line 311: nit: typo in encryption, but I'd probably just do "Unencrypted or Per-User Encrypted Home Directories"

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -951,10 +956,139 @@ Step 6: First Boot
    Replace ``YOUR_USERNAME`` with your desired username::
 
      username=YOUR_USERNAME
+     
+   Choose one of the following:
+     
+   - Unencrypted homedir or whole-disk encryption
```
rlaager, line 962: `Unencrypted, ZFS native encryption, or LUKS:`

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -951,10 +956,139 @@ Step 6: First Boot
...
+     
+       ::
+
+         zfs create rpool/home/$username
+         adduser $username
+       
+   - Encrypted homdir per user, automatic decryption on login
```
rlaager, line 969: `Per-user Encrypted Home Directories:`

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -951,10 +956,139 @@ Step 6: First Boot
...
+     
+       ::
+       
+         zfs create rpool/home/$username -o encryption=on -o keyformat=passphrase -o keylocation=prompt -o canmount=noauto -o org.openzfs.systemd:ignore=on
+         zfs mount rpool/home/$username
+         adduser $username
+         
+       **Note**: Use the same strong password for ZFS encryption and user password. Please note: After a breach of your password changing the ZFS password does not restore protection.
```
rlaager, line 977: nit: Add a comma after "After a breach of your password"

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -951,10 +956,139 @@ Step 6: First Boot
...
+       
+       Tell PAM to unlock the dataset key on login::
+       
+         printf "auth [success=1 default=ignore] pam_succeed_if.so service = sudo quiet\n\
```
rlaager, line 981: I'd probably make this use `vi` as the example. I've tried to make the longer stuff just invoke the editor. I think I used to use multi-line heredocs and people didn't love that.

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -951,10 +956,139 @@ Step 6: First Boot
...
+                 auth optional pam_exec.so expose_authtok /usr/local/sbin/unlock-key-zfs-homedir\n"\
+           >> /etc/pam.d/common-auth
+           
+       The second line calls a new unlock-script, the first line disables it if we just sudo.
```
rlaager, line 985: Why do we want to disable it if we just sudo? I'm not saying you're wrong, but can you explain it a little more. To me in comments here is fine. The explanation doesn't have to be in the document itself at this point.

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -951,10 +956,139 @@ Step 6: First Boot
...
+       
+       Create the unlock-script::
+       
+         touch /usr/local/sbin/unlock-key-zfs-homedir 
+         chmod a+x /usr/local/sbin/unlock-key-zfs-homedir 
+         vi /usr/local/sbin/unlock-key-zfs-homedir 
+       
+       and put into it::
```
rlaager, line 993: 1. shellcheck on this
2. I don't see a reason to require bash over plain sh.
3. Is there a particular reason we couldn't allow for an encrypted root home directory?

I get this instead:
```
#!/bin/sh

set -e

if [ "$PAM_USER" = "root" ]
then
    dataset=rpool/root
    mountpoint=/root
else
    dataset=rpool/home/$PAM_USER
    mountpoint=/home/$PAM_USER
fi

# Exit if already mounted.
findmnt "$mountpoint" && exit 0

# If the dataset does not exist, this will exit.
encryption=$(zfs list "$dataset" -o encryption -H 2> /dev/null) || exit 0

[ "$encryption" = off ] && exit 0

zfs load-key "rpool/home/$PAM_USER" < /dev/stdin
```

I'm not sure if that `< /dev/stdin` is accomplishing anything. Wouldn't stdin be connected to `zfs load-key` by default?

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -951,10 +956,139 @@ Step 6: First Boot
...
+         # exit if already mounted for some reason
+         findmnt "/home/$PAM_USER" && exit 0
+
+         # still here? unlock now
+         zfs load-key "rpool/home/$PAM_USER" < /dev/stdin
+         
+       Now create a systemd service to mount our unlocked dataset::
+       
+         mkdir -p /etc/systemd/system/user@.service.d/
+         printf "[Unit]\n\
```
rlaager, line 1020: Same thing here. Use `vi`.

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -951,10 +956,139 @@ Step 6: First Boot
...
+       
+         printf "auth [success=1 default=ignore] pam_succeed_if.so service = sudo quiet\n\
+                 auth optional pam_exec.so expose_authtok /usr/local/sbin/unlock-key-zfs-homedir\n"\
+           >> /etc/pam.d/common-auth
+           
+       The second line calls a new unlock-script, the first line disables it if we just sudo.
+       
+       Create the unlock-script::
+       
+         touch /usr/local/sbin/unlock-key-zfs-homedir
```
rlaager, line 989: Instead of `touch`, `chmod`, `vi`, I would just `vi` then `chmod`. Same below.

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
```diff
@@ -951,10 +956,139 @@ Step 6: First Boot
...
+         
+       Finally, create the helper script::
+       
+         touch /usr/local/sbin/mount-zfs-homedir
+         chmod a+x /usr/local/sbin/mount-zfs-homedir
+         vi /usr/local/sbin/mount-zfs-homedir
+         
+       And put into it::
+       
+         #!/bin/bash
```
rlaager, line 1053: Same general comments as the script above.

rlaager (changes requested):
I've made a first pass through this. Most of this is minor nitpicking. I haven't had a chance to actually walk through the steps yet.

rlaager:
Am I understanding the theory of operation correctly here? When you login, PAM passes your password to the unlock script. The unlock script does the `zfs load-key`. Then the system unit mounts your home directory?

At logout, systemd unmounts the home directory and locks it?

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
christophhagemann: Systemd doesn't start a user session when using sudo. There is no point in unlocking our dataset if no one is going to mount it.

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
christophhagemann: I think it's not wise to allow for an encrypted root home directory, because (see above) systemd doesn't start a user session when using sudo, so root would end up without mounted home directory.

christophhagemann:
Am Montag, dem 30.01.2023 um 13:51 -0800 schrieb Richard Laager:
> Am I understanding the theory of operation correctly here? When you
> login, PAM passes your password to the unlock script. The unlock
> script does the zfs load-key. Then the system unit mounts your home
> directory?
> At logout, systemd unmounts the home directory and locks it?

Almost. PAM runs the unlock script on every login. On subsequent logins
the script does nothing. The systemd mount script on the other hand is
only run on first login and last logout to avoid errors when other
sessions are still running.

One corner case is sudo. Systemd seems to ignore sudo: no user session
is started when using sudo -u, so sudo -u <youruser> -i ends up without
mounted home directory. I don't think this is a common use case except
for root. This is why I chose not to support /root/ encryption.

docs/Getting Started/Debian/Debian Bullseye Root on ZFS.rst
christophhagemann: BTW: I was pretty surprised `zfs load-key` worked without explicit `< /dev/stdin`.

ne9z:
I don't mean to pour cold water on this, just my two cents.

I think that any customization procedures that can be done post-installation -- any that does not involve repartitioning the disks -- should be done post-installation and documented elsewhere, like a blog or a wiki.  The guide here should only be a starting point.

The installation guides maintained by myself previously suffered from feature-creep, I added Secure Boot, encrypted boot pool, SSH remote unlock, among others.  The main disadvantages are the following:

- The procedures are not specific to the distro, and could be reused if it's posted on a wiki instead, such as Arch Linux wiki.
- Whenever the instructions need update, maintainers at openzfs/openzfs-docs must be involved, which increases their workload.  In a wiki, they can be updated easily by a volunteer.
- The burden of testing every possible combination of configuration documented in the guide are shouldered by us, which is quite time-consuming.
- The extra options make the guide longer and more difficult to follow.

Therefore I moved the instructions to Arch Linux wiki, Alpine wiki, Fedora wiki and NixOS wiki respectively and dropped them from this repo.  We now benefit from the oversight of volunteers there, an easier-to-follow installation guide, and less maintenance work.  In particular, the NixOS Root on ZFS guide only requires 8 minutes(!) to follow for a multi-disk encrypted installation, excluding time used for downloading live media.

christophhagemann:
> I think that any customization procedures that can be done post-
> installation -- any that does not involve repartitioning the disks --
> should be done post-installation and documented elsewhere, like a
> blog or a wiki. The guide here should only be a starting point.

I'm just a user. For me it would have been helpful to find this info in
one place. It's not strictly post-install IMHO.

>  * The procedures are not specific to the distro

Good point. May I suggest to move it to a separate page below "Getting
started" - "Making use zfs' features" or similar and linking there from
the distro-specific pages? 

>  * Whenever the instructions need update, maintainers at
> openzfs/openzfs-docs must be involved, which increases their
> workload.
>  * The burden of testing every possible combination of configuration
> documented in the guide are shouldered by us, which is quite time-
> consuming.

I'd volunteer to maintain this page.

>  * The extra options make the guide longer and more difficult to
> follow.

Would be solved by creating a separate page.

If there's consensus a separate distro-independent page would be
desirable, I'd invest the time to rework my contribution and maintain
the page afterwards.

gmelikov:
Personally, I like universal small blocks of info. Have such pages on distros' wiki is good, but it requires to have contributors there. 

So, LGTM to have dedicated pages here for these purposes, main thing to have appropriate navigation position for them. Maybe @rlaager or @ne9z have other opinions and may challenge me here.

grahamperrin:
I don't suppose that anything here helps, but FYI (merged): 

- https://github.com/freebsd/freebsd-src/pull/881

Issue #371, opened 2022-12-27
Can't boot if single drive missing with mirrored pools


nasamaher:
Hi,

A mirrored install should allow for booting even if one of the drives is offline, right?

I've tried installing multiple times (without errors) on mirrored 4GB SSDs using UEFI and get the same results.  Both drives must be online.  If one drive is disconnected I have a problem.

lsblk output after the install:
![image](https://user-images.githubusercontent.com/1706941/209701187-c6a3f476-d60e-4161-84cd-cd15f69db809.png)

If I remove /dev/sda, then I get the Grub menu, but upon trying to load the OS I get

_"Unknown device 0.  You need to load the kernel first"_

If I remove /dev/sdb, then on boot, no Grub menu; I get dropped to Grub Shell (without an apparent warning message)

I've gotten the same result twice.  I should note in both installs I also was faced with not being able to import rpool, but I see this is a common issue and I resolved it in the initramfs shell.

prehonor:
installing the boot loader...
setting up /etc...
/etc/tmpfiles.d/journal-nocow.conf:26: Failed to resolve specifier: uninitialized /etc detected, skipping
All rules containing unresolvable specifiers will be skipped.
updating GRUB 2 menu...
installing the GRUB 2 boot loader on /dev/disk/by-id/ata-WDC_WD10EZEX-08WN4A0_WD-WCC6Y5YF14H1...
Installing for i386-pc platform.
Installation finished. No error reported.
installing the GRUB 2 EFI boot loader into /dev/disk/by-id/ata-WDC_WD10EZEX-08WN4A0_WD-WCC6Y5YF14H1-part1...
Installing for x86_64-efi platform.
/nix/store/plikx1pi79sf46hbwz73416bd9865lwi-grub-2.06/sbin/grub-install: error: failed to get canonical path of `devtmpfs'.
/nix/store/cxzmkzazap2337gx1dfb6mk3kmdhrj2c-install-grub.pl: installation of GRUB EFI into /dev/disk/by-id/ata-WDC_WD10EZEX-08WN4A0_WD-WCC6Y5YF14H1-part1 failed: No such file or directory

It failed again,saddly

gmelikov:
@ne9z wasn't it fixed by https://github.com/openzfs/openzfs-docs/pull/384 ?

ne9z:
This issue was filed in December 2022, and yesterday, another person
commented on this issue for a totally unrelated problem.  I do not
[...]

PR #386, opened 2023-02-10
NixOS Root on ZFS: use Nix-config for multi-disk

ne9z:
NixOS has enjoyed popularity among ZFS users thanks to its declarative configuration and native ZFS support. However, the installation guide used hardcoded disk names in configuration files, which is unnecessary and is the source of difficulties in multidisk setups.

The guide is now rewritten to leverage expressions in the Nix language to manage multidisk setups.

@gmelikov 


ne9z:
@gmelikov  Completed 3 test-installations for both single and mirrored setups.  I think it's ready for merging now.  I was also unable to reproduce the issue reported in #371, I'll contact the reporter to test if the issue is fixed with this new guide.

gmelikov (approved):
Thank you! We'll see how it works to have artifacts nearly with rsts.

PR #389, opened 2023-02-16
Fix missing options to accompany comments

m3t:
In my eyes the comments related to X11 should be removed or else be amended by option examples.
At least `xkbOptions` has to enclose the broken statement.

### DIFF
--- a/docs/Getting Started/NixOS/Root on ZFS/configuration-immutable.nix
+++ b/docs/Getting Started/NixOS/Root on ZFS/configuration-immutable.nix
@@ -44,9 +44,9 @@ in {
   #   useXkbConfig = true; # use xkbOptions in tty.
   # };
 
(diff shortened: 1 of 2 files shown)


ne9z:
Thanks for the feedback!

Yes, the default config hints should be kept in sync with the config
template generated by nixos-generate-config:

    https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/installer/tools/tools.nix#L124

Can you also update the configuration-immutable.nix file, so that

    diff configuration.nix configuration-immutable.nix

only outputs the differences necessary for implementing immutable root
filesystem?

Thanks again!

ne9z (approved):

ne9z:
@gmelikov LGTM.  Approved.

gmelikov (approved):
Thank you!

PR #390, opened 2023-02-17
Add FreeBSD packages to the build instructions

allanjude:


docs/Developer Resources/Building ZFS.rst
```diff
@@ -56,6 +56,12 @@ The following dependencies should be installed to build the latest ZFS
 
    sudo apt install build-essential autoconf automake libtool gawk alien fakeroot dkms libblkid-dev uuid-dev libudev-dev libssl-dev zlib1g-dev libaio-dev libattr1-dev libelf-dev linux-headers-generic python3 python3-dev python3-setuptools python3-cffi libffi-dev python3-packaging git libcurl4-openssl-dev
 
+-  **FreeBSD**:
+
+.. code:: sh
+
+   pkg install autoconf automake autotools git gmake python devel/py-sysctl sudo
```
gmelikov, line 63: It's not mandatory to have sudo on Linux to build packages without test runs, just curious why do we need it on FreeBSD.

allanjude: it isn't required, but is for the tests, and many instructions contain a sudo prefix

gmelikov (approved):

PR #392, opened 2023-02-22
Workload Tuning: actualize volblocksize

gmelikov:
As someone in Twitter (c) pointed, it was a little bit obsolete and obscure.


gmelikov:
@shipilev thank you for pointing out obsolete docs, feel free to comment on new version.

shipilev:
> @shipilev thank you for pointing out obsolete docs, feel free to comment on new version.

Thank you for handling this.

"guest FS" instead of "overlay FS" maybe?

Still not clear what to do with the ubiquitous case like putting the ext4 filesystem on top of zvol. AFAIK, ext4 cannot do block size larger than the hardware page size, so it is pretty much never more than 4K on current x86_64 hosts. Would setting `volblocksize` to `4K` be sensible there? Do we recommend it? Or do we still recommend running with 8K or 16K `volblocksize`, incurring some write amplification?

It might suffice to say: "Current default balances the metadata overhead and compression opportunities, while incurring some write amplification on overlay FSes that run with smaller block sizes. Users are advised to test their scenarios and see whether the zvol block size needs to be changed to favor one or the other: <list follows>"

gmelikov:
@shipilev thank you, I've got your point and like your wording, expanded motivation and added useful source links.

PR #399, opened 2023-03-16
Update NixOS configuration, clean up RHEL, Fedora, Arch guide

### LINKED ISSUE #397: RHEL Root Documentation errors
The latest version of the instructions for RHEL Root on ZFS has a few errors there are several redundant commands.  The issues are noted below:

"Preparation" Section 11: the command
 `dnf install -y gdisk dosfstools arch-install-scripts`
fails as the arch-install scripts is missing dependencies.  The next two commands download and install the arch-install-scripts anyway.  I would suggest changing the arch-install-scripts in the dnf  command to cryptsetup instead as that will solve the next issue.

System Installation Section 1:  This section fails to complete as cryptsetup is not installed by default in the latest Alma Linux GNOME Mini Live boot image.

"System Installation" Section 4:  The following lines will install the wrong kernel version in the new /mnt partition when using the newest Alma Linux live Image due to the repositories having a newer kernel available.  You have to install the same kernel as is running on the live image or there will be issues when dracut builds initramfs.
`dnf --installroot=/mnt \
--releasever=$VERSION_ID -y install \
@core  grub2-efi-x64 \
grub2-pc-modules grub2-efi-x64-modules \
shim-x64  efibootmgr \
kernel`

To solve this change to:

`dnf --installroot=/mnt \
--releasever=$VERSION_ID -y install \
@core  grub2-efi-x64 \
grub2-pc-modules grub2-efi-x64-modules \
shim-x64  efibootmgr \
kernel-$(uname -r)`

"System Installation" Section 4: missing commands for mounting the /var, /var/log, and /var/lib partitions

Also at the end of Section 4 the command
`zfs create -o mountpoint=legacy     rpool/alma/root`
should be changed to 
`zfs create -o mountpoint=/ -o canmount=noauto rpool/alma/root`
as the next line 
`mount -t zfs -o zfsutil rpool/alma/root /mnt`
will fail with the current version.  This is also redundant as the root partition is unmounted, altered to match the recommended creation syntax, down further, then instructions are given to issue
`mount -t zfs rpool/alma/root /mnt`
Which, again, fails because it only works with mountpoint=legacy

I have not made not of the redundant commands that affect functionality, but those should probably be cleaned up as well to make the procedure more concise.

ne9z:
This pull request upgrades NixOS configuration to be Flake-based, and cleans up ZFS dataset creation section in RHEL, Fedora and Arch.

Thanks @DrArclight

Closes #397


ne9z:
@gmelikov

ne9z:
@rlaager Could you review and merge this?

gmelikov:
@ne9z I'm a little bit confused when you tag me but push other changes for some time after that (sometimes even several days in a row), not a big problem, but I started to postpone my review because of that (to get at least half a day since last push), to be honest. 

Again - not a big problem, but I propose to make self-sufficient ready-to-merge commits, and give them to review. I try not to affect your contributions with review process, but it would be great to be on the same page here.

I'm ready to review it as before, but if you want to speed up merge process - let's make than straightforward.

docs/Getting Started/Alpine Linux/Root on ZFS/2-system-installation.rst
```diff
@@ -40,23 +40,18 @@ System Installation
...
-     hole_birth
-     large_blocks
-     lz4_compress
-     spacemap_histogram
-     EOF
-
-      zpool create \
-          -o compatibility=/root/grub2 \
+      zpool create -d \
+          -o feature@async_destroy=enabled \
```
gmelikov, line 44: What do you think about using https://github.com/openzfs/zfs/pull/11818 ?

ne9z: It is not possible to use compatibility= option on Alpine, because Alpine does not package the necessary files [source](https://pkgs.alpinelinux.org/contents?branch=edge&name=zfs&arch=aarch64&repo=main).

ne9z:
My bad.  I'm a bit of a perfectionist in this regard -- I can always find something wrong with my own guide and be tempted to change them.

ne9z:
It's done for real now, this time.  Please proceed with review.

docs/Getting Started/Alpine Linux/Root on ZFS/2-system-installation.rst
```diff
@@ -269,6 +257,8 @@ Post installaion
 
      setup-desktop
 
+#. Configure swap.
```
gmelikov, line 260: Don't we need to have some commands here? Or it was intended just to notify user about right place?

ne9z: Based on the principle that _"if not specific to this Root on ZFS document, look elsewhere"_, the user is advised to perform their own configuration of the swap partition.  Such as dmcrypt, cryptsetup with key file, or encryption with /dev/random, or plain unencrypted swap partition, along with related parameters.

This is already the case for Arch, Fedora and RHEL.
Previously Alpine guide did not even inform the user that the swap partition was left unconfigured, so this PR adds that notice.

gmelikov (approved):
Thank you!

PR #396, opened 2023-03-13
Build instructions for native Debian packages

usaleem-ix:
This commit updates the build instructions for native Debian packages.

PR #400, opened 2023-03-16
[openzfs/zfs#14611] Fix missing install from source steps.

aamkye:
I was missing this info, so I'm glad to provide them :)
Please let me know if certain information is missing.

### DIFF
--- a/docs/Developer Resources/User friendly, zero to hero, build from source till working guide.rst
+++ b/docs/Developer Resources/User friendly, zero to hero, build from source till working guide.rst
@@ -1,0 +1,243 @@
+User friendly, zero to hero, build from source till working guide (Ubuntu based)
+============
+
+GitHub Repositories
+~~~~~~~~~~~~~~~~~~~
+
+The official source for OpenZFS is maintained at GitHub by the
+`openzfs <https://github.com/openzfs/>`__ organization. The primary
+git repository for the project is the `zfs
+<https://github.com/openzfs/zfs>`__ repository.
+
+Installing Dependencies
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The first thing you'll need to do is prepare your environment by
+installing a full development tool chain. In addition, development
+headers for both the kernel and the following packages must be
+available. It is important to note that if the development kernel
+headers for the currently running kernel aren't installed, the modules
+won't compile properly.
+
+The following dependencies should be installed to build the latest ZFS release.
+
+-  **Debian, Ubuntu**:
+
+.. code:: sh
+
+   sudo apt install build-essential autoconf automake libtool gawk alien fakeroot dkms libblkid-dev uuid-dev libudev-dev libssl-dev zlib1g-dev libaio-dev libattr1-dev libelf-dev linux-headers-generic python3 python3-dev python3-setuptools python3-cffi libffi-dev python3-packaging git libcurl4-openssl-dev
+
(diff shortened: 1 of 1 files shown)


aamkye:
This "tutorial" is based on my PoC and HomeLab installation via Ansible tested on ubuntu server 22.10.

aamkye:
Will need to update docs according to #396

gmelikov:
Why don't we just add part about services (and other small nits) into Building ZFS.rst as additional section? Looks like a duplicate otherwise, we already have 2 instructions about building from source.

aamkye:
So maybe we should merge them into one covering all details?

aamkye:
@gmelikov I'll prepare something (merged doc) according to the docs and my knowledge.

gmelikov:
@aamkye yes, I've meant that we may just add needed things to existing page. 

Side note - while it may be handy, I don't think it would be great to add parts about ansible. We may describe ansible somewhere else and add a link to it (maybe whole playbook example as yaml file? But we'll need to maintain it).

aamkye:
Valid point Sir! So let's stick to bash commands and bash commands only in this repo :)

gmelikov:
Just to be on the same page, we already have:
- https://openzfs.github.io/openzfs-docs/Developer%20Resources/Custom%20Packages.html
- https://openzfs.github.io/openzfs-docs/Developer%20Resources/Building%20ZFS.html

I think the best place is https://openzfs.github.io/openzfs-docs/Developer%20Resources/Custom%20Packages.html , because the point is to get working custom packages, not just build modules.

aamkye:
@gmelikov I've also found that besides missing services there are also a missing cron jobs: `scrub` and `trim` and scripts:
- `/usr/lib/zfs-linux/trim`
- `/usr/lib/zfs-linux/scrub`

I think it also should be part of the documentation, shouldn't it?

gmelikov:
@aamkye in fact, there are 2 types of packages in any project:
- produced by project itself - they are usually don't integrate in every aspect of specific distro
- produced by distro's package maintainers - these packages may have additional integrations and patches

So, we may describe explicitly about such difference, and that our documentation speaks about first type. That mean is that (for example in Debian) if you want to get 100% identical package to original distro's one - you need to rebuild from distro's source package, not project's one.

If project's package misses something - we may discuss to add missing parts in our packages.

So, any of additions in docs may describe just notes about which functionality may be added, but we should now describe them as "must haves" (because in this case we should add them into packages, not into documentation).

You may be interested to read this thread with Debian's package maintainer https://github.com/openzfs/zfs/pull/13451

gmelikov:
Yep, that's exactly what I'm talking about - every Linux distribution may have it's own package with differences and patches. We even can't describe uniform deb info because Ubuntu, Debian and Proxmox all have their own packages.

aamkye:
So maybe as a ZFS community, we could gather people from different distros and prepare a repository to prepare such packaging for each distribution containing guidelines, tutorials, and scripts to do so not only for the ZFS project but for more. Because it seems I've entered a rabbit hole and it goes waaaaay deeper than I thought.

useribs (changes requested):
As of {2025-01} [ZFS](https://openzfs.github.io/openzfs-docs/Developer%20Resources/Custom%20Packages.html) is  also missing dh-dkms package. Without it and and libcurl4-openssl-dev it does not build

PR #401, opened 2023-03-18
NixOS: enclose disk variable in double quotes

ne9z:
Just discovered another bug.  This will cause multi-disk setups to be treated as the same as single disk, with all other disks except the first one ignored.  @gmelikov


ne9z:
Sorry to bother you again!  I hope this would be the last bug for this week.

gmelikov (approved):

gmelikov:
@ne9z don't worry, appreciate your work!

PR #409, opened 2023-03-25
Clarify that NixOS cfg is needed to make zfs.ko available

nrdvana:
The documented configuration changes for NixOS appear to be making zfs available for boot.  However, these changes are also required just to make the zfs.ko module available to modprobe even for users who don't need ZFS available at boot time.  Also, the kernel module does not appear until after a reboot, regardless of 'nixos-rebuild switch'.

(a more knowledgable NixOS user might know how to modprobe
 without a reboot, but I don't)

### DIFF
--- a/docs/Getting Started/NixOS/index.rst
+++ b/docs/Getting Started/NixOS/index.rst
@@ -30,6 +30,10 @@ see below.
 
 Live image ships with ZFS support by default.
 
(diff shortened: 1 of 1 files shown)


ne9z (approved):

gmelikov (approved):

gmelikov:
Thank you!

PR #410, opened 2023-03-31
NixOS, Alpine Linux: mount datasets in Root on ZFS guide; immutable root; GRUB update

### LINKED ISSUE #398: alpine root on zfs guide missing set root password or login via authorized key before exit chroot/reboot
on [openzfs-docs](https://openzfs.github.io/openzfs-docs/Getting%20Started/Alpine%20Linux/Root%20on%20ZFS/2-system-installation.html) at item 17 its unmount the filesystem, before this though, 
setting the root password is required to be able to login after the reboot (item 18).

kinda minor issue since its just the irritating sequence of chrooting via the liveusb just to set the password,
before getting into the system.

another solution could also be to setup the sshd before rebooting and setup login via authorized pubkey.

ne9z:
Acknowledgement: @bjornstromberg
Closes #398


docs/Getting Started/NixOS/Root on ZFS/1-preparation.rst
```diff
@@ -17,6 +17,12 @@ Example, Tow-Boot is installed to an SD card.  Then the SD card
 should not be also shared with NixOS.  Install NixOS to an external
 disk instead.
 
+**Note for Immutable Root**
+
+You can enable immutable root by setting ``my.boot.immutable`` option
+inside the flake to ``true``.  Set to ``false`` reverts to
+non-immutable root file system.  Defaults to ``false``.
```
SuperSandro2000, line 24: This is not something that upstream NixOS supports.

ne9z:
Upstream also does not support Root on ZFS at all.  At least I was
unable to find any such option in the installer shipped in the live
media.  Does this fact render the whole guide obsolete?  It does not, of
course.

In the case of other distros, Debian, Ubuntu, Fedora, RHEL, Alpine, it
is precisely because the lack of upstream support that the guides here
were written.

For your information, if you look inside the repo to see what that
``my.boot.immutable`` option does, it only changes how the datasets are
mounted, nothing fancy:

https://github.com/ne9z/dotfiles-flake/blob/058bef759b26c1d37bfb4998f532db8358e71da7/modules/boot/default.nix#L88

I do not think dropping this feature, which is very popular among root
on ZFS users, is a good idea.

SuperSandro2000:
> Upstream also does not support Root on ZFS at all.

It does, I am running it on multiple machines and I am writting this comment from a ZFS on root machine.

> At least I was unable to find any such option in the installer shipped in the live media.

The installer does not have ZFS support but you can manually configure the disk like before.

> For your information, if you look inside the repo to see what that ``my.boot.immutable`` option does, it only changes how the datasets are mounted, nothing fancy:

This is not something upstream NixOS supports or tests for. This could just fall apart in the future because it is not accounted for when tests are run.

ne9z:
Sandro ***@***.***> writes:

>> Upstream also does not support Root on ZFS at all.
>
> It does, I am running it on multiple machines and I am writting this
> comment from a ZFS on root machine.

Immutable root is supported by upstream because I am running it on my
three laptops and home server and I am writting this comment from a
immutable ZFS on root machine for over two years and four releases.

No less valid than your logic.

>> At least I was unable to find any such option in the installer
>> shipped in the live media.
>
> The installer does not have ZFS support but you can manually configure
> the disk like before.

So what is your definition of upstream Root on ZFS support?  I see that,
the guide as it is now, runs on my machine, so it is supported.

>> For your information, if you look inside the repo to see what that
>> ``my.boot.immutable`` option does, it only changes how the datasets
>> are mounted, nothing fancy:
>
> This is not something upstream NixOS supports or tests for. This could
> just fall apart in the future because it is not accounted for when
> tests are run.

This option, ``my.boot.immutable``, directly, exclusively and only uses
config.fileSystems.* options.  Does NixOS run tests for
config.fileSystems.* options?

I guess it does, because fileSystems.* options are absolutely critical.

Now, I think we should keep the flake-based approach.  You want
something clean like:

my.boot.devices = [ "disk1" "disk2" ]

or do you like

fileSystems = {
  /boot/efis/disk1 = {
    device = /dev/disk/by-id/disk1;
  };
  /boot/efis/disk2 = {
    device = /dev/disk/by-id/disk2;
  };
}

grub.mirroredBoots = { [ "disk1" "disk2" ] }


Because we must support and correctly handle multi-disk installations.
This is what ZFS is all about.

ne9z:
If you agree that we should keep the modules and flake, then we can
discuss the other issues you mentioned in the other PR.

> NixOS allows using multiple programs to manage your network settings, so
> not everyone necessarily has network manager installed.

Please read what is inside the cloned repo before commenting.


NetworkManager is enabled by default for all users who has cloned the
repo.  This is done out of consideration of wireless network users.  If
not enabled, wireless network is not possible with minimal NixOS
installation profile.

If you just go ahead and making comments without even looking at what
you are talking about, it would be a waste of time for both of us.

> Immutable root is not an upstream setting and it shouldn&#39;t be
> recommended in the basic ZFS install guide.

Discussed earlier.

ne9z:
To keep discussion in one place, let's address your other concerns from
issue #411:

> The NixOS guide right now guides people to clone a dotfiles repo and
> change some settings in there, instead of explaining what are the
> relevant settings people should set in their NixOS configuration.

I have replied there that the main draw backs are the following:

-  The configuration will be no longer reproducible.

-  ARM64 support will be dropped.

-  Immutable root support will be dropped.

-  SSH remote encryption will be dropped.

-  Managing multiple hosts in single repo will be dropped.

-  Managing multi-disk system will require verbose configuration without
   custom module.

> Setting up ZFS on NixOS is pretty easy.

Just because you got Root on ZFS installed on your laptop does not mean
your approach meets all the requirements of a Root on ZFS documentation.
For example, do you have multi-disk and encrypted swap configured? With
mirrored bootloader?

Mind you, without utilizing disk arrays, ZFS does not provide any
redundancy.

Once the single disk in your pool fails, your data is gone
for good.

> and IMO cloning some extra config should not be required, as I believe
> that the guide could fit on one screen in standard text size when
> leaving out the explanations.

It would not fit in one screen, given all the goals we have set out to
accomplish.

> There are a handful of module options be taught about and recommended
> to enable (autoScrub, autoTrim, etc)

This part should be delegated to NixOS wiki.  All configuration steps
which do not involve repartitioning the disks, changing the pool
properties or changing the dataset layout is out-of-scope for this
guide.

ne9z:
> instead of explaining what are the relevant settings people should
> set in their NixOS configuration.

We will provide detailed and extensive comments in the flake repo of
each custom module and option enabled to provide a guide for users
configuring their own system.

The comments are currently missing, because I do not have the time at
the moment.  Contributions from the community is much appreciated.

gmelikov (approved):
LGTM from a formatting side, @ne9z feel free to ping me if it's ready to merge.

ne9z:
@gmelikov It's now ready for merging.  PR #412 and #262 can also be closed.  The latter is no longer relevant to the current RHEL and Fedora guide.

To @SuperSandro2000: I have removed NetworkManager on your suggestion.

ne9z:
I just received a feedback (#413) about Arch Linux, where the initrd defaults was changed upstream such that the guide no longer boots.  Fix added to this PR.

PR #412, opened 2023-04-01
nixos: remove network manager, immutable root

SuperSandro2000:
NixOS allows using multiple programs to manage your network settings, so not everyone necessarily has network manager installed. Immutable root is not an upstream setting and it shouldn't be recommended in the basic ZFS install guide.

### DIFF
--- a/docs/Getting Started/NixOS/Root on ZFS/3-system-configuration.rst
+++ b/docs/Getting Started/NixOS/Root on ZFS/3-system-configuration.rst
@@ -91,22 +91,6 @@ System Configuration
 
      reboot
 
-#. NetworkManager is enabled by default.  To manage network
-   connections, execute::
-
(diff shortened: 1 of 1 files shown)


ne9z:
Thanks for the pull request.  However, as long as the config in that
repo is not changed, it is meaningless to only remove them here.  There
will still be an immutable option in the per-host configuration and
NetworkManager will still be installed by default.

Please do not open issues and pull requests on the same and related
issue multiple times.  We should keep the discussion in one place, then
open a single pull request once consensus is reached.

ne9z:
NetworkManager is already removed in #410.  I have no intention to remove immutable root as it is widely used by NixOS users.  If you don't like it, don't use it.  It is optional and disabled by default.

Close as WONTFIX.

PR #421, opened 2023-04-06
NixOS: add CI/CD tests for Root on ZFS guide

ne9z:
The guide was updated with literate programming techniques, such that a executable program can be extracted from the guide itself and can be tested.  [Here is the script generated from reStructuredText documentation](https://github.com/ne9z/personal-dotfiles/blob/6a7f7c8fad52e89e037ac63d74d699ba40f0c0a6/install.sh)
This allows us to enable continuous tests on the guide and therefore improve reliability.

However this involves some additional configuration of Github Actions, which must be reviewed.

Also, there is the difficulty that KVM accelerated virtualization is [not supported on free accounts](https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/) thus the task require a long time to run.  But I have a spare machine that can be set up as a self-hosted runner for this purpose.  On my machine with KVM acceleration enabled, the test takes about 15 minutes to complete.

Or should we just run tests locally instead of running them as Actions?

Marked as draft for discussion.

@gmelikov


ne9z:
After much consideration, I think it is better to keep the tests in another repo.  I will open another pull request only for changes in the guide itself.

gmelikov:
@ne9z interesting idea! Some thoughts on topic:
- I'm ok to have tests here, we want to run them on PRs, after all
- I'd try to stick with github's runners until we can, just to minimize our dependencies
- looks like macos's runners have nested virtualization, but it may be too costly to implement it, as example https://github.com/vmactions/freebsd-vm

How long did it run in github's runner, after all? Did it succeed in 6h (default limit)?

ne9z:
Bad news and good news.

Bad news is that it didn't finish in 6h, with qemu virtualization.

Good news is that, I just discovered the Github Runners, being based on Ubuntu latest 22.04, supports ZFS natively.  I can do all the tests I need in a chroot!  Such a pleasant surprise.  I just need to replace qemu commands with chroot, and that was it!

The test will create two blank disk images and mount them as loopback devices.  The test script then create all the necessary partitions and ZFS pools on those loopback devices.

The test takes about 5 minutes to finish. In chroot.

ne9z:
I think I can finish writing and testing the CI/CD tests today or tomorrow.

gmelikov:
@ne9z great! Just wanted to note, that we use Ubuntu runners for our code tests extensively https://github.com/openzfs/zfs/tree/master/.github/workflows ,maybe it'll be useful. It even has docker https://github.com/openzfs/zfs/blob/master/.github/workflows/checkstyle.yaml#L46

PR #422, opened 2023-04-07
NixOS: update guide to be compatible with automated tests.

ne9z:
Also remove unmaintained Arch Linux guides.

This PR only contains changes for the guide itself.  Tests are now in [a separate repo](https://github.com/ne9z/nixos-live) and can be executed locally.


ne9z:
Chroot test is available now.  See #421.  Closing.

PR #423, opened 2023-04-08
Alpine, Arch Linux, Fedora, RHEL, NixOS Root on ZFS: add CI/CD tests

ne9z:
Also move parts of Arch Linux guide to Arch Linux wiki and added Boot Environment page with CI/CD test coverage.

All tests finish successfully, as can be seen [here](https://github.com/ne9z/openzfs-docs/actions/runs/4647024578/jobs/8223800977).

Phew! That was a lot of work.  Hopefully this effort can bring the reliability of the guides to a new level.

PR #425, opened 2023-04-18
fix dracut config in root on ZFS docs

zeromind:
step 3 incorrectly uses "forced_drivers";
the dracut config option is "force_drivers" (force, not force*d*)

see https://github.com/search?q=repo%3Adracutdevs%2Fdracut+force_drivers&type=code


ne9z:
Thanks and applied onto PR #425

    https://github.com/openzfs/openzfs-docs/pull/423

You can view the upcoming, simplified guide here:

    https://github.com/ne9z/openzfs-docs/blob/master/docs/Getting%20Started/RHEL-based%20distro/Root%20on%20ZFS.rst

Makes me wonder why dracut does not emit any warning or errors about a
invalid config.

gmelikov:
Thank you, merged in [#423](https://github.com/openzfs/openzfs-docs/pull/423)

PR #435, opened 2023-04-30
Update Building ZFS.rst

alek-p:
update configure tips to assume "ZFS build" and use more consistent wording


gmelikov (approved):
Thank you!

PR #438, opened 2023-05-16
Arch, Alpine, Fedora, RHEL: update to latest system version

ghost:
- Update RHEL version from 9.1 to 9.2, fix broken download link.
- Update Fedora version from 37 to 38
- Update Arch version from 20230401 to 20230503
- Update Alpine version from 3.17 to 3.18
- Update Fedora, RHEL zfs-repo from 2-2 to 2-3
- Fix one typo in NixOS docs
- Add two lines of missing commands in zfs_root_maintenance.rst

Closes #434

@gmelikov

### DIFF
--- a/docs/Getting Started/zfs_root_maintenance.rst
+++ b/docs/Getting Started/zfs_root_maintenance.rst
@@ -143,7 +143,6 @@ and other maintenance operations and is thus currently not supported.
       boot_dataset=$(df -P /boot | tail -n1 | cut -f1 -d' ' || true )
       root_dataset=$(df -P / | tail -n1 | cut -f1 -d' ' || true )
       new_boot_dataset="${boot_dataset%/*}"/"${my_boot_env}"
-      boot_dataset=$(df -P /boot | tail -n1 | cut -f1 -d' ' || true )
       rm_boot_dataset=$(head -n1 new_boot_env_entry_"${new_boot_dataset##*/}" | sed 's|^# *||' || true )
 
       if [ "${boot_dataset}" = "${rm_boot_dataset}" ]; then


docs/Getting Started/zfs_root_maintenance.rst
```diff
@@ -140,6 +140,8 @@ and other maintenance operations and is thus currently not supported.
    ::
 
       set -x
+      boot_dataset=$(df -P /boot | tail -n1 | cut -f1 -d' ' || true )
```
gmelikov, line 143: Looks like it already is here, line 146 in new variant.

ne9z:
Yes.  But someone who wish to destroy the BE would very likely miss the
step where boot_dataset variable was set, which is currently at the very
beginning of BE creation steps.  So, we mention it again inside the BE
destruction steps to avoid any confusion.

docs/Getting Started/zfs_root_maintenance.rst
gmelikov: Then let's just remove second declaration from line 146?

gmelikov (approved):
Thank you!

ne9z:
Thanks for reviewing and merging this!  I see that the Sphinx build
failed somehow; and the pages were not updated after merging this PR to
master.  Maybe we should trigger a manual rebuild of the GitHub pages.

gmelikov:
@ne9z yep, http://cvsweb.netbsd.org/bsdweb.cgi/~checkout~/src/external/cddl/osnet/dist/cmd/zpool/zpool-features.7?content-type=text/plain&only_with_tag=MAIN... doesn't work now, I'll rerun CI when it will be available.

PR #440, opened 2023-05-30
Bump links to 2-3 also

quartsize:
7ecad4c (Bump EL RPMs to ver 2-3, 2023-04-28) updated the yum/dnf-install snippets.

I tried to find a way to DRY it out, but I was unable to find a way to make ReST do a substitution inside another directive.


tonyhutter (approved):
Good catch!  Thank you for this fix.

PR #448, opened 2023-07-23
Install ZFS in the Live CD environment in multiple steps to avoid errors

llacqie:
When installing zfsutils-linux on Debian it installs before zfs-dkms, but when installing it tries to modprobe zfs-dkms module, which is not yet available. So it would be nice to install zfs-dkms first, and only then zfsutils-linux.


JonChesterfield:
I just ran into this a couple of years later (symptom was no zfs module loaded), though I'm inclined to think this is a bug in Debian rather than in the docs. I worked around by installing linux-headers then dpkg-reconfigure zfs-dkms, as suggested in https://github.com/openzfs/zfs/issues/13061

Issue #447, opened 2023-07-14
Could we keep versioned instances of the man pages at least around?


rincebrain:
Sometimes we do things like delete fields or parameters in newer releases, and it's sometimes useful to point people to those without having to dig through git history.

(This specifically came to mind when reading openzfs/zfs#15052 and thinking "we should maybe keep the references to explain the old fields around" and then going "no keeping old versions of the docs for each release series is probably a better solution")

gmelikov:
Done 
![image](https://github.com/openzfs/openzfs-docs/assets/3868786/0e76aed0-3e14-4e6b-8388-bac87bd9f441)

PR #450, opened 2023-07-29
Man pages: generate for different major versions

gmelikov:
Closes https://github.com/openzfs/openzfs-docs/issues/447

TODO:
- [X] redirect from previous urls


behlendorf:
@gmelikov very nice!  Thanks for tackling this, being able to quickly lookup, and directly link to, the matching version of the documentation is great.

PR #453, opened 2023-09-11
Update index.rst

tomchiverton:
Add optional step to configure Fedora to never remove zfs, even if it means keeping an older kernel

https://github.com/openzfs/zfs/issues/15258


tomchiverton:
I'm only using ZFS for /home etc - not root, so can't test it, but there's no reason it wouldn't apply there too?

tonyhutter (approved):

gmelikov (approved):
Thank you!

PR #444, opened 2023-06-22
Initial Debian bookworm Root on ZFS page

jim-perkins:
Initial Doc for Debian 12 Bookworm root on ZFS.  Pretty much what was discussed with a few other things noticed when I verified the page.  Let me kow if there is anything else needed.  thanks so much for the existing documentation.

#442 
@rlaager

PR #454, opened 2023-09-13
Stop udev before zpool export

stackcoder:
In my setup systemd-udevd reproducible locks and prevent exporting the zpool with: `pool is busy`.

After stopping udev I was able to export the zpool. However shutting it down before rebooting the whole machine should not make it worse. Also much nicer than booting into an "dirty" zpool within initramfs.


stackcoder:
I can also reproduce this behavior on Debian Bookwork. Upcoming doc here: #444

gmelikov:
I've tried it lately, in my case it wasn't udev

stackcoder:
@gmelikov You might share what it caused in your case? Maybe this also works for me.

gmelikov:
@stackcoder unfortunately I didn't have time to investigate further, I've just manually imported pools with force on first boot.

PR #460, opened 2023-10-18
Use compatibility property while creating boot pool

colmbuckley:
Since OpenZFS 2.1, the `compatibility` property has allowed users to specify groups of pool features, to maintain compatibility with specific systems or versions. The `grub2` feature set is designed for use with GRUB2 boot pools, so can be used here replacing a bunch of distinct feature settings.


gmelikov (approved):
LGTM, checked manually before/after PR flags, they're nearly the same functionality-wise.

PR #465, opened 2023-10-26
Root on ZFS: Warn against native encryption; add NixOS tutorial for LUKS

ne9z:
Hopefully I've got the reStructuredText indentation right.  This PR contains two changes:

- Add warning against ZFS native encryption, with references
- Add LUKS-based guide for NixOS

Tested with https://github.com/ne9z/openzfs-docs/actions/runs/6651356857/job/18073228694

### DIFF
--- a/docs/Getting Started/Alpine Linux/Root on ZFS.rst
+++ b/docs/Getting Started/Alpine Linux/Root on ZFS.rst
@@ -352,7 +352,7 @@ System Installation
 
    - Encrypted:
 
-     Pick a strong password. Once compromised, changing password will not keep your
+     Avoid ZFS send/recv when using native encryption, see `a ZFS developer's comment on this issue`__ and `this spreadsheet of bugs`__.    A LUKS-based guide has yet to be written. Once compromised, changing password will not keep your
      data safe. See ``zfs-change-key(8)`` for more info
 
      .. code-block:: sh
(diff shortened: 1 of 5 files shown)


docs/Getting Started/Alpine Linux/Root on ZFS.rst
```diff
@@ -340,7 +340,7 @@ System Installation
 
    - Encrypted:
 
-     Pick a strong password. Once compromised, changing password will not keep your
+     ZFS native encryption is buggy, see `a ZFS developer's comment on this issue`__ and `this spreadsheet of bugs`__.  In short, if you care about your data, don't use native encryption.  A LUKS-based guide has yet to be written. Once compromised, changing password will not keep your
```
gmelikov, line 343: While you MAY have problems mainly with zfs send with some arguments, on default case it's pretty well working. I use native encryption on my laptop at least for 2-3 years now. I propose to add more specific warning. What do you think about:

```suggestion
     While ZFS native encryption is a production ready feature, it has some corner cases, see `a ZFS developer's comment on this issue`__ and `this spreadsheet of bugs`__.  Once compromised, changing password will not keep your
```

ne9z:
Well, I'm not too certain about the "production ready" phrasing.  To
quote rincebrain again:

> Depending on which problem, sometimes this is "just" a kernel panic,
> sometimes it mangles your key settings so you need something custom
> and magic to let you reach in and fix it, sometimes it writes records
> that should not have been allowed in an encrypted dataset and then
> errors out trying to read them again. (To pick three examples.)

What do you think?

ne9z:
Quote from ElvishJerricco:

> One of those bugs even leaked plaintext on disk (#14330)

Quote from rincebrain in that issue

> So it seems like somehow we generated an embedded write record on an
> encrypted dataset. Whoopsie.

So at this point, not even the promise of proper encryption has been
fulfilled by ZFS native encryption.  You might consider this a
disadvantage.

ne9z:
@gmelikov I don't know what your intentions are.  Should we hide the fact that native encryption codebase is unmaintained and buggy?  In any case, I have updated the pull request to address your comments above.

gmelikov:
@ne9z of course we should not hide problems, but if something so terribly broken in stable releases, then we have to disable it in code at all, or at least escalate it in code repo. Plus this is an official documentation, we should be careful with (un)ambiguous declarations.

I like your wording, thank you!

PR #466, opened 2023-10-26
Root on ZFS: use compat=legacy with boot pool

ne9z:
As mentioned [here](https://github.com/openzfs/openzfs-docs/issues/463#issuecomment-1775955088), compat=legacy seems to fix the problem where, once a snapshot of the root level dataset of the boot pool is taken, boot pool will become irreversibly unreadable by GRUB.

Also added a note against using less well-tested features in general.

### DIFF
--- a/docs/Getting Started/Alpine Linux/Root on ZFS.rst
+++ b/docs/Getting Started/Alpine Linux/Root on ZFS.rst
@@ -40,6 +40,10 @@ to their site for installation details.
 Unless stated otherwise, it is not recommended to customize system
 configuration before reboot.
 
+**Only use well-tested pool features**
+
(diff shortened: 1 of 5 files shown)


gmelikov:
legacy is too strict, but why not, at least temporarily. We may add a note about grub bug in future, if upstream won't fix it in a year.

colmbuckley:
If people are following this guide, bpool *needs* very few zpool features in order to remain useful, we should definitely prune `compatibility.d/grub2` to only contain features we are highly confident to work.

ne9z:
Well, as a student of mathematics, only a proof would make me confident
about anything.  Since we do not have that, I will have to contend with
legacy.

To be clear, there is no guarantee (proof) that legacy will work with
GRUB either.

PR #467, opened 2023-10-26
Root on ZFS: mention ZFSBootMenu

ne9z:
@ahesford @zdykstra two main developers of ZBM:  I hope I have written a neutral description of ZBM.  Kudos for developing it.

### DIFF
--- a/docs/Getting Started/Alpine Linux/Root on ZFS.rst
+++ b/docs/Getting Started/Alpine Linux/Root on ZFS.rst
@@ -30,7 +30,9 @@ of ZFS features on the boot pool. [In general, bootloader treat disks
 as read-only to minimize the risk of damaging on-disk data.]
 
 `ZFSBootMenu <https://zfsbootmenu.org>`__ is an alternative bootloader
-free of such limitations and has support for boot environments.  Refer
+free of such limitations and has support for boot environments. Do not
+follow instructions on this page if you plan to use ZBM,
+as the layouts are not compatible.  Refer
 to their site for installation details.
 
 **Customization**
--- a/docs/Getting Started/Alpine Linux/Root on ZFS.rst
+++ b/docs/Getting Started/Alpine Linux/Root on ZFS.rst
@@ -22,6 +22,16 @@ Alpine Linux Root on ZFS
     apk update
     apk add alpine-conf curl
 
+**ZFSBootMenu**
+
+This tutorial is based on the GRUB bootloader.  Due to its independent
+implementation of a read-only ZFS driver, GRUB only supports a subset
+of ZFS features on the boot pool. [In general, bootloader treat disks
+as read-only to minimize the risk of damaging on-disk data.]
+
+`ZFSBootMenu <https://zfsbootmenu.org>`__ is an alternative bootloader
+free of such limitations and has support for boot environments.  Refer
+to their site for installation details.
 
 **Customization**
 
--- a/docs/Getting Started/Arch Linux/Root on ZFS.rst
+++ b/docs/Getting Started/Arch Linux/Root on ZFS.rst
@@ -35,7 +35,9 @@ of ZFS features on the boot pool. [In general, bootloader treat disks
 as read-only to minimize the risk of damaging on-disk data.]
 
 `ZFSBootMenu <https://zfsbootmenu.org>`__ is an alternative bootloader
-free of such limitations and has support for boot environments.  Refer
+free of such limitations and has support for boot environments. Do not
+follow instructions on this page if you plan to use ZBM,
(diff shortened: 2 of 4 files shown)


docs/Getting Started/Alpine Linux/Root on ZFS.rst
```diff
@@ -22,6 +22,16 @@ Alpine Linux Root on ZFS
...
 
+**ZFSBootMenu**
+
+This tutorial is based on the GRUB bootloader.  Due to its independent
+implementation of a read-only ZFS driver, GRUB only supports a subset
+of ZFS features on the boot pool. [In general, bootloader treat disks
+as read-only to minimize the risk of damaging on-disk data.]
+
+`ZFSBootMenu <https://zfsbootmenu.org>`__ is an alternative bootloader
+free of such limitations and has support for boot environments.  Refer
```
gmelikov, line 33: IIUC, zbm supports only zfs. If that's so, I propose to distinct them by this factor mainly. Like "zbm is a more feature-rich and zfs-problems-proof, but it's a zfs-only thing, while grub is a usual default choice which can help with multiboot etc".

ne9z:
Good idea.

docs/Getting Started/Alpine Linux/Root on ZFS.rst
ahesford: ZFSBootMenu is not EFI-only.

ahesford:
I don't really see the value of these additions. Not only is ZFSBootMenu an alternative to GRUB, but it won't tolerate having kernels on a separate filesystem from the root. (Our views on separate boot pools are abundantly clear.) A primary purpose of these guides seems to be to show users how to install distributions with a separate pool for `/boot`, making them completely moot for users who want to adopt ZFSBootMenu.

If you want to leave a reference to ZFSBootMenu, please change your last sentence ("Refer to their site for installation details.") to something like
> However, the filesystem layout proposed in this guide will be incompatible with ZFSBootMenu. Alternative installation guides for several distributions are provided in the ZFSBootMenu documentation.

ne9z:
"Andrew J. Hesford" ***@***.***> writes:

> If you want to leave a reference to ZFSBootMenu, please change your
> last sentence ("Refer to their site for installation details.") to
> something like
>
>  However, the filesystem layout proposed in this guide will be
>  incompatible with ZFSBootMenu. Alternative installation guides for
>  several distributions are provided in the ZFSBootMenu documentation.

Looks good to me.

docs/Getting Started/Alpine Linux/Root on ZFS.rst
gmelikov: @ahesford could you point me at some docs about usage without efi? I've looked at docs, specifically debian manual (because I'm most familiar with it) https://docs.zfsbootmenu.org/en/v2.2.x/guides/debian.html and it doesn't have mbr examples, so I thought about efi as main vector of usage. I've found only one example about mbr https://docs.zfsbootmenu.org/en/v2.2.x/search.html?q=Mbr&check_keywords=yes&area=default

My point is - if we link zbm, then we can highlight main use cases of zbm to make it easier for reader to choose.

zdykstra: @gmelikov The primary installation example for it is at https://docs.zfsbootmenu.org/en/v2.2.x/guides/void-linux/syslinux-mbr.html# .  I don't think it's unreasonable to say ZFSBootMenu is primarily used on EFI systems, but `syslinux` does provide a convenient way to boot it as well.

ahesford: The key take-away here is that ZFSBootMenu can be built either as a single UEFI executable (what the world has taken to calling a "UKI") or as separate kernel and initramfs pairs. We ship pre-built versions of each with every release as a convenience; users can also create their own custom versions using `generate-zbm` or our build containers. While the UEFI executable can be booted directly, the separate kernel and initramfs can be booted using a standard bootloader (gummiboot, syslinux, refind, etc.) that is capable of launching a Linux kernel and giving it an initramfs. Just put the ZBM kernel and initramfs on some file system (EFI system partition, dedicated ext4 partition, or the like) and point the bootloader where it needs to go.

gmelikov: @ahesford @zdykstra thanks for clarifications! I assume that you're prepared for people without EFI then.

(Looks like even grub may load ZBM, afrer all :) )

On topic - I'm ok to leave it as is in PR with @ahesford 's note about ending.

gmelikov (approved):

PR #483, opened 2024-01-19
Update distribution version, switch to rEFInd and systemd-boot

### LINKED ISSUE #411: NixOS: mention relevant settings instead of using dotfiles repo
The NixOS guide right now guides people to clone a dotfiles repo (https://github.com/openzfs/openzfs-docs/blob/master/docs/Getting%20Started/NixOS/Root%20on%20ZFS/3-system-configuration.rst) and change some settings in there, instead of explaining what are the relevant settings people should set in their NixOS configuration.

Setting up ZFS on NixOS is pretty easy and IMO cloning some extra config should not be required, as I believe that the guide could fit on one screen in standard text size when leaving out the explanations.
There are a handful of module options (https://search.nixos.org/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=zfs) which people should be taught about and recommended to enable (autoScrub, autoTrim, etc)

ne9z:
- Switch from GRUB to rEFInd and (when possible) systemd-boot.  This is in sync with upstream changes.
- For NixOS, description for relevant options are added.
- Update distribution version.
- Reduced total length of instruction by 1100 lines.

Closes #411.


ne9z:
@gmelikov I think this is ready for review.  NetBSD link seems to be failing.

gmelikov (approved):
LGTM

PR #485, opened 2024-01-26
Fedora Root on ZFS: Add a link to Greg\'s installation script

gregory-lee-bartholomew:
People might find this script useful.

### DIFF
--- a/docs/Getting Started/Fedora/Root on ZFS.rst
+++ b/docs/Getting Started/Fedora/Root on ZFS.rst
@@ -28,6 +28,11 @@
 Fedora Root on ZFS
 =======================================
 
+Notes
+~~~~~
+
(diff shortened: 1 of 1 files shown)


gmelikov:
I'm not against such notes, but let's make them uniform, what do you think to make it same as in https://openzfs.github.io/openzfs-docs/Getting%20Started/openSUSE/openSUSE%20Leap%20Root%20on%20ZFS.html#notes ? It's important to make notes that it's an unofficial tool.

Another note - @ne9z made https://github.com/openzfs/openzfs-docs/blob/master/docs/Getting%20Started/Fedora/Root%20on%20ZFS.rst in a way that it may be at least tested in CI (so, commands can me merged in a script), if we mention other tools, then we should describe their differences.

gmelikov:
I propose to add a link as a note in https://github.com/openzfs/openzfs-docs/blob/master/docs/Getting%20Started/Fedora/Root%20on%20ZFS.rst .

CI is not needed here at all, I've just wanted to describe our existing docs :) If you want to run some tests - it may be better to run them in your repository.

gregory-lee-bartholomew:
@gmelikov: Is this now closer to what you have in mind? I'm happy to make further changes if you see anything that could use improvement. Just let me know. 🙂

gmelikov (approved):
Thank you!

PR #486, opened 2024-01-28
Fix for Custom Packages page, packages for Debian\Ubuntu

Dante4:
Fix for native-deb-utils, it requires libpam0g-dev, otherwise dpkg-checkbuilddeps fails with error unmet build dependencies

The build was tested with currently released 2.2.2

photo_2024-01-28_01-22-55


gmelikov (approved):
Thank you!

PR #488, opened 2024-01-29
Fixed same typo in the Debian Bullseye and Bookworm docs.

DebianDoesDallas:
In the root on ZFS docs for both Debian Bullseye, and Bookworm, there is a typo where what should be "initramfs" reads "initamfs".


gmelikov (approved):
Thank you!

PR #489, opened 2024-01-29
Add "parallel" to Debian/Ubuntu build dependencies

Fabian-Gruenbichler:
there's another variant from "moreutils" that doesn't support `-X0` which is used for `make checkstyle`. 

I'll file an issue for improving the detection there as well.


gmelikov (approved):
Thank you!

PR #492, opened 2024-02-09
fix: avoid unintended bold text in `Building ZFS.rst`

dresslerdemos:
In RST, spacing and indentation are significant, and consistent usage is essential for the correct parsing of the document.

Because the first line in the adjusted bullet points were differently indented than the following lines, the "boldness" of the first characters was extended to the whole line.

I noticed this by chance on [Building ZFS](https://openzfs.github.io/openzfs-docs/Developer%20Resources/Building%20ZFS.html) and fixed it in this one document. I did not review other pages for similar mistakes.


gmelikov (approved):
Thank you!

PR #498, opened 2024-03-02
Debian packages: specify dracut usefulness

### LINKED ISSUE #497: Custom Packages instructions for Debian/Ubuntu will never work
As of this writing:
```
To build RPM converted Debian packages:

$ cd zfs
$ ./configure --enable-systemd
$ make -j1 deb-utils deb-kmod
$ sudo apt-get install --fix-missing ./*.deb

Starting from openzfs-2.2 release, native Debian packages can be built as follows:

$ cd zfs
$ ./configure
$ make native-deb-utils native-deb-kmod
$ rm ../openzfs-zfs-dkms_*.deb
$ sudo apt-get install --fix-missing ../*.deb
```

However, this is going to emit both the `dracut` and `initramfs` helper packages, and in at least Ubuntu, both of those Provide a common thing and Conflict with each other, so if it attempts to install both to satisfy that, it'll fail, as demonstrated in openzfs/zfs#15948.

I've personally used something like:
`sudo apt install $(ls *VERSION*deb | grep -v dracut | while read LINE;do echo -n "./${LINE} ";done;)`

since 2018 for this, looking at my shell history. 

It's probably easiest to just advise people not to install the dracut packages by default, though, with a note explaining how to do the opposite if they're using a not-default initrd generator.

(Opening a bug because I'm busy and don't have time to go through the insane clone-edit-PR loop we turned simply editing a wiki into.)

gmelikov:
Closes #497

### DIFF
--- a/docs/Developer Resources/Custom Packages.rst
+++ b/docs/Developer Resources/Custom Packages.rst
@@ -179,7 +179,7 @@ as follows:
    $ ./configure
    $ make native-deb-utils native-deb-kmod
    $ rm ../openzfs-zfs-dkms_*.deb
-   $ rm ./openzfs-zfs-dracut_*.deb  # deb-based systems usually use initramfs
+   $ rm ../openzfs-zfs-dracut_*.deb  # deb-based systems usually use initramfs
    $ sudo apt-get install --fix-missing ../*.deb
 
 Native Debian packages build with pre-configured paths for Debian and
@@ -211,7 +211,7 @@ packages can be built as follows:
    $ cd zfs
    $ ./configure
    $ make native-deb-utils
-   $ rm ./openzfs-zfs-dracut_*.deb  # deb-based systems usually use initramfs
+   $ rm ../openzfs-zfs-dracut_*.deb  # deb-based systems usually use initramfs
    $ sudo apt-get install --fix-missing ../*.deb
 
 Get the Source Code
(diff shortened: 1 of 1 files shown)


docs/Developer Resources/Custom Packages.rst
```diff
@@ -210,6 +211,7 @@ packages can be built as follows:
    $ cd zfs
    $ ./configure
    $ make native-deb-utils
+   $ rm ./openzfs-zfs-dracut_*.deb  # deb-based systems usually use initramfs
```
rincebrain, line 214: You want ../ here, it emits packages in the parent dir, which is why the install command on the next like is `../*.deb`

docs/Developer Resources/Custom Packages.rst
```diff
@@ -179,6 +179,7 @@ as follows:
    $ ./configure
    $ make native-deb-utils native-deb-kmod
    $ rm ../openzfs-zfs-dkms_*.deb
+   $ rm ./openzfs-zfs-dracut_*.deb  # deb-based systems usually use initramfs
```
rincebrain, line 182: Same as the other comment.

docs/Developer Resources/Custom Packages.rst
gmelikov: You're right, my sleepy eyes were too closed, that's why I adore reviews :)

PR #502, opened 2024-04-08
performance: fix innodb configuration variable name

elelayan:
The variable is innodb_doublewrite, while the command-line option is --skip-innodb-doublewrite

See https://mariadb.com/docs/server/ref/mdb/system-variables/innodb_doublewrite/


gmelikov:
I may be wrong (don't have time to check it fast), but same docs says that 

> innodb_doublewrite - Enable InnoDB doublewrite buffer (enabled by default). Disable with **--skip-innodb-doublewrite**.

gmelikov (approved):
Thank you!

Checked on `Server version: 10.3.39-MariaDB-0+deb10u2 Debian 10`, both variants work well for me, but let's use more straightforward way.

elelayan:
thanks for the quick merge :-)

PR #505, opened 2024-04-19
Hardware.rst: Mention Kingston DC1000B

nh2:
I've been using this one with ZFS in my laptop for years, with it being one of the only M.2 devices that are short enough for laptops while still having power loss protection.

Source: https://www.kingston.com/en/ssd/dc1000b-data-center-boot-ssd

### DIFF
--- a/docs/Performance and Tuning/Hardware.rst
+++ b/docs/Performance and Tuning/Hardware.rst
@@ -505,6 +505,7 @@ follows:
 
 -  Intel 750
 -  Intel DC P3500/P3600/P3608/P3700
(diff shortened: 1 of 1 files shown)


nh2:
Some newer M.2 2280 devices with power loss protection I found in research just now:

* https://europe.kioxia.com/en-europe/business/ssd/client-ssd/xg6-p.html
* https://www.transcend-info.com/embedded/product/embedded-ssd-solutions/mte712p

gmelikov (approved):
Thank you!

Issue #508, opened 2024-05-27
Can someone please decipher this?


systemcrash:
https://github.com/openzfs/openzfs-docs/blob/55a0d1cf73b564a889a3fb616fddf55be0f33f14/docs/Basic%20Concepts/RAIDZ.rst#write

Please elaborate. This smells like a Russian thought it and wrote it roughly in English.

What is "full stripe width"? If you made an assumption of what "full" means - please define it. 

What is "one block write"? Abstract, singular? What on earth does "has a write IOPS of one slowest disk because of that" mean?

"of one slowest disk"? Because of "that" what is "that" here?

If I am to take a guess (which a user should not be doing, because the user reads the documentation in order that they not guess) - were you trying to say something like the following?:

```
A stripe spans all drives in the array. A one block write will write the stripe part onto each disk.
A RAIDZ vdev has a write IOPS of the slowest disk in the array in the worst case because writes to each disk must be completed before successive writes on the other, faster disks can commence.
```

gmelikov:
> Please elaborate. This smells like a Russian thought it and wrote it roughly in English.

You're right, I've made it as a draft, and it existed far too long in this state.

I hope to rework it someday, https://www.reddit.com/r/zfs/comments/ujba3i/does_allocation_overhead_still_exist_with/ looks like a well-worded example to use.

Feel free to send PRs if you're interested, I'd be glad to review them!

systemcrash:
Hello :)

The link is generally regarding compression, while the essence of the text in this issue is about write performance. The linked text might be added under a separate heading, so.... let's tackle the issue first.

Does my proposed text capture the essence of what you intended to say?

gmelikov:
@systemcrash feel free to review me, thank you https://github.com/openzfs/openzfs-docs/pull/509/files

PR #509, opened 2024-06-03
RAIDZ: rephrase `write` section

gmelikov:
Closes: #508

### DIFF
--- a/docs/Basic Concepts/RAIDZ.rst
+++ b/docs/Basic Concepts/RAIDZ.rst
@@ -87,5 +87,5 @@ Performance considerations
 Write
 ^^^^^
 
-Because of full stripe width, one block write will write stripe part on each disk.
-One RAIDZ vdev has a write IOPS of one slowest disk because of that in worst case.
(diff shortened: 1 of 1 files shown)


docs/Basic Concepts/RAIDZ.rst
```diff
@@ -87,5 +87,5 @@ Performance considerations
 Write
 ^^^^^
 
-Because of full stripe width, one block write will write stripe part on each disk.
-One RAIDZ vdev has a write IOPS of one slowest disk because of that in worst case.
+A stripe spans across all drives in the array. A one block write will write the stripe part onto each disk.
+A RAIDZ vdev has a write IOPS of the slowest disk in the array in the worst case because write operation of all stripe parts must be completed on each disk.
```
systemcrash, line 91: `A RAIDZ vdev has a write IOPS of the slowest disk in the array in the worst case because the write operation of all stripe parts must be completed on each disk.`

systemcrash (changes requested):

PR #518, opened 2024-08-07
Improve Root on ZFS

upsj:
The installed base system is missing kernel modules that would otherwise be installed in a regular Rocky installation, especially network drivers, leaving the resulting system in a hard-to-use state.

Additionally, we shouldn't use HTTP to download a file without checking its integrity.


gmelikov (approved):
Thank you!

PR #520, opened 2024-08-18
Update Fedora docs

reavessm:
Update docs to only install the kernel-devel package that matches the running kernel.  This is specifically needed when running in a live environment where kernel upgrades are not persisted after a reboot.

This was tested with the F40 Workstation installer


gmelikov (approved):
Thank you, it won't hurt

PR #522, opened 2024-08-29
Fix default value of zio_taskq_batch_pct

0mp:


docs/Performance and Tuning/Module Parameters.rst
```diff
@@ -6294,7 +6294,7 @@ therefore require recompiling the kernel module to adjust.
 | Range               | 1 to 100, fractional number of CPUs are       |
 |                     | rounded down                                  |
 +---------------------+-----------------------------------------------+
-| Default             | 75                                            |
+| Default             | 80                                            |
```
gmelikov, line 6297: Thank you!

I'll propose to add info about version changes https://github.com/openzfs/zfs/commit/7457b024ba2be2cf742e07239c20a1c3f3fa9c72 , something like :

```suggestion
| Default             | 75 for versions before v.2.2.0, 80 after      |
```

Issue #457, opened 2023-10-13
Ubuntu: Build dependencies missing libpam0g-dev and dh-dkms


jaredjones:
As of `zfs-2.2`, the Ubuntu required dependencies are missing `libpam0g-dev` and `dh-dksm` which results in the following error when building the project. The OpenZFS documentation should be updated to include this dependency.
```
Unmet build dependencies: libpam0g-dev
```

Note: `dh-dkms` is required as of Ubuntu Mantic to run: `make native-deb-utils` as its no longer part of `dkms` proper.

rnickle:
I can confirm that those packages are required for Debian 12 and Ubuntu 24.04.

Also, I observed for Ubuntu 24.04 the following packages were required, I suspect that they may have been pulled in as dependencies by other packages for Debian 12 because attempting to install them found they were already present:

- libblkid-dev
- dh-autoreconf
- libtirpc-dev

Here is what I tested for Debian 12 and Ubuntu 24.04 for _make native-deb_ and _make native-deb-utils_:

```shell
$ sudo apt install alien autoconf automake build-essential debhelper-compat dh-autoreconf dh-dkms dh-python dkms fakeroot gawk git libaio-dev libattr1-dev libblkid-dev libcurl4-openssl-dev libelf-dev libffi-dev libpam0g-dev libssl-dev libtirpc-dev libtool libudev-dev linux-headers-generic parallel po-debconf python3 python3-all-dev python3-cffi python3-dev python3-packaging python3-setuptools python3-sphinx uuid-dev zlib1g-dev
```

PR #524, opened 2024-09-13
Add missing packages for Debian derived systems

rnickle:
While building for Ubuntu 24.04, I documented that five packages were missing that were required for different aspects of configuration and building of ZFS, associated utilities, packaging and libraries.

I found that the _libpam0g-dev_ and _dh-dkms_ packages were clearly required by Debian 12 and Ubuntu 24.04 and not present on my provisioned installs.  In general, I believe all 5 packages are required on both platforms, although the different platforms appear to pull in some packages from other dependencies.

Specifically, these packages appear to be required for both Debian 12 and Ubuntu 24.04:

- dh-autoreconf
- dh-dkms
- libblkid-dev
- libpam0g-dev
- libtirpc-dev

As validation, I performed the build in parallel on a newly provisioned Ubuntu 24.04 VM and a newly provisioned Debian 12 VM.

I was specifically interested in building _native-deb_ and _native-deb-utils_ so the package list reflects that requirement.

 I followed these steps for Debian and Ubuntu:

```shell
$ sh autogen.sh
$ ./configure
$ make -s -j$(nproc)
$ make deb-native
$ make deb-native-utils
```
Incidentally fixed a typo.

Issues:

#457

### DIFF
--- a/docs/Developer Resources/Building ZFS.rst
+++ b/docs/Developer Resources/Building ZFS.rst
@@ -143,7 +143,7 @@ the ZFS and SPL source in the traditional autotools fashion.
 | **tip:** Native Debian packages build with pre-configured paths for
   Debian and Ubuntu. It's best not to override the paths during
   configure.
-| **tip:** For native Debain packages, ``KVERS``, ``KSRC`` and ``KOBJ``
+| **tip:** For native Debian packages, ``KVERS``, ``KSRC`` and ``KOBJ``
   environment variables can be exported to specify the kernel installed
   in non-default location.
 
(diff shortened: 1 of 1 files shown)


gmelikov:
I've checked with our CI https://github.com/openzfs/zfs/blob/b1958b531b75afa8a138eae8ace9d17201fa42b9/.github/workflows/scripts/qemu-3-deps.sh#L32

and some of new packages are absent there:
- libtirpc-dev
- dh-autoreconf
- dh-dkms

, BUT native-deb is a little bit new here and CI don't build them IIRC, so I don't see a problem to add them.

Thank you!

PR #532, opened 2024-11-28
Update custom packages docs for Fedora/RHEL

gcoretech:
@AllKind asked me to update the docs for Fedora in https://github.com/openzfs/zfs/issues/16807#issuecomment-2500530385

I've installed Fedora 41 Workstation in a virtual machine and checked which rpm packages are required to successfully build zfs (2.2.6.tar.gz and git 2.3.99). I've also added a paragraph about UEFI/secure boot which is common with Fedora. Just ignore the commit if you don't want that published.

### DIFF
--- a/docs/Developer Resources/Custom Packages.rst
+++ b/docs/Developer Resources/Custom Packages.rst
@@ -154,6 +154,57 @@ option must be passed to configure.
    $ make -j1 rpm-utils rpm-kmod
    $ sudo dnf install *.$(uname -m).rpm *.noarch.rpm
 
+Fedora 41 secure boot with kmod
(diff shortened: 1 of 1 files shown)


AllKind:
Wow, that was fast! Thank you very much!

gmelikov (approved):
Thank you! LGTM, unfortunately I don't have Fedora now to easily test, but it's not a stopper.

tonyhutter (approved):
Thanks for the doc updates, this looks fine.  Let's not merge it until zfs-2.2.7 comes out first though (https://github.com/openzfs/zfs/pull/16781) - that will be the first release that supports Fedora 41.

PR #538, opened 2025-04-02
Fix link to Aaron Toponce's tutorial

vog:
It seems that Aaron Toponce's tutorial is no longer online. So let's link to the latest archived version on archive.org instead.

Technical details: While the domain pthree.org still resolves to a proper IP address, as well an IPv4 address, there appears to be no webserver behind it to respond, neither on port 443 nor on legacy port 80.


gmelikov (approved):
Thank you! That's a pity that it went dark. 

We may want to introduce some actual alternative for this quide, but it'll be another story.

PR #539, opened 2025-04-13
Rocky Linux Root on ZFS: downgrade Alpine Linux to 3.18

CodeZap27:
Using Alpline Linux 3.19, Rocky Linux would fail during startup due to Alpine Linux having a newer zfs kmod version than the zfs tools version in Rocky Linux


gmelikov (approved):
Thank you!

Issue #500, opened 2024-03-23
Debian Bookworm Root on ZFS guide not working- cannot export rpool.


chrcoluk:
The pool cant be exported at end of its initial setup. 

Lose-lose situation, reboot anyway, but first boot will fail due to the lack of an export, have to import manually on initramfs prompt.

Or find whatever is causing the lock on the rpool, however I was unable to do this, I suspect the cause is the addition of installing the timesync service on the bookworm guide, as I have followed the guide 5 times.

First time I didnt install the timesync service, and export worked fine as was the case on the bullseye guide.

All other 4 occasions I installed the timesync service and try as I might I could not export the pool, the 5th attempt is when I tried to carry on with brute force without the export, although I dont know if the lack of export will have caused data corruption.

Dont think I have the motivation to go through the guide again a 6th time right now, but perhaps should have tried shutting down the service prior to the exit of chroot, then I suspect the rest will go without hiccup.

denppa:
In my case, this error did occur, and I upon reboot, I just had to import rpool in initramfs prompt. It is not ideal, but a good enough work around.

Issue #353, opened 2022-10-06
Ubuntu 22.04 Root on zfs - cannot export 'rpool': pool is busy


lwysocki:
Hi, I have issue as below at step 5.10. I use Live CD for 22.04.1.

![image](https://user-images.githubusercontent.com/26348598/194249222-3cb2c818-4c20-42c1-bdae-3d814953b25a.png)

lwysocki:
So after reboot I imported rpool by:
```
zpool import -d /dev/disk/by-id/{rpool-partition}
zpool import -f rpool
exit
```

nasamaher:
I had the same problem over mutliple attempts.  I suspected it was because there was a lingering ~ubuntu login that was hanging on to rpool.  I never got it to work.  Boot fails and down in grub command line I did something similar

`zpool import -f rpool`
`zpool export rpool`

Then booting worked.

PR #540, opened 2025-04-13
Debian+Ubuntu root-on-zfs: add a last-resort command to export pool

gmelikov:
There may be orphan processes after chroot, so try to kill them.

With that I could get 100% reproducible export when I build cloud images (they have to boot first time successfully, otherwise it's not a ready-to use root-on-zfs image):
- https://github.com/infraguys/gci_ubuntu_zfs
- specifically - https://github.com/infraguys/gci_ubuntu_zfs/blob/master/templates/ubuntu-24/scripts/prepare-zfs-root.sh#L103-L119


gmelikov:
Closes #500 #353 #270

PR #541, opened 2025-04-22
Add donations page

tonyhutter:
Add a link to our donations page on SPI.


robn (approved):
Reads fine, links work. I wouldn't hate it if there was a bit more about what we spend it on, but nbd. Thanks!

gmelikov (approved):
LGTM

PR #544, opened 2025-05-13
Update Ubuntu 22.04 Root on ZFS for Raspberry Pi.rst

jimklimov:
Clarify what we prepare as a "partition script" and why -- how it fits into the bigger picture.


jimklimov:
Probably makes sense to replicate this text to other OS versions.

It may be reasonable to hint for people doing this en-masse (for flashing numerous machines) that they can dump their prepared disk after removing partition 3 and before expanding the ZFS pool (the part of disk until the sector where partition 3 used to start), so they have an image that can be immediately flashed to other storage units (maybe with some first-boot script injected to rename UUIDs of the machine-id, rootfs, SSH keys, etc.).

rlaager:
Thanks!

Issue #550, opened 2025-05-28
RHEL-Based Distro: Update Online Documentation to Warn Users OpenZFS is Unfit for Production Use on RHEL Systems


ghost:
The <a href="https://openzfs.github.io/openzfs-docs/Getting%20Started/RHEL-based%20distro/index.html">online documentation</a> is out of date and does not reflect the current state of development of OpenZFS. As evident by the fact the <a href="https://github.com/openzfs/zfs/issues/17332">kernel module does not even build on the non testing branch of OpenZFS</a>, <a href="https://github.com/openzfs/zfs/issues/17364">the 2.2.x branch has an unresolved issue on EL9.6 kernels</a>, and <a href="https://github.com/openzfs/zfs/issues/17290#issuecomment-2842289398">concerns regarding stability of the current stable branches</a>, OpenZFS is not suitable for production environments on RHEL and RHEL based platforms, and this should be clearly noted in the documentation.

As the documentation is <a href="https://github.com/openzfs/zfs/issues/17221#issuecomment-2800081691">"totally hand-made"</a>, I am more than happy to contribute the necessary changes so development effort may focus on more critical issues.

PR #551, opened 2025-05-28
Include temporary warning of stability on RHEL platforms

ghost:
Update documentation to include warning that ZFS is currently unfit for production use on RHEL systems. Closes https://github.com/openzfs/openzfs-docs/issues/550

### DIFF
--- a/docs/Getting Started/RHEL-based distro/index.rst
+++ b/docs/Getting Started/RHEL-based distro/index.rst
@@ -1,6 +1,14 @@
 RHEL-based distro
 =======================
 
+Important Notice Regarding Stability on RHEL-based Distros
+-----------------------------------------------------------
+
+As of May 2025, we are currently experiencing stability and build issues on this platform. 
+The kernel modules may not build on non-testing branches, and known regressions exist in 
+the 2.2.x branch on EL9.6 kernels. **We strongly advise against deploying OpenZFS in production
+environments on these platforms until these issues are resolved.**
+
 Contents
 --------
 .. toctree::


namazso:
Not unfit, unsupported: https://github.com/openzfs/zfs/issues/17360#issuecomment-2898019707

macthecadillac:
From the docs, it is quite clear that OpenZFS is often behind on supporting new EL minor releases. In fact, if you try the command given in the documentation
```bash
dnf list available --showduplicates --releasever=9.6 kmod-zfs
```
or go to the repository at http://download.zfsonlinux.org/epel/, it is obvious that version 9.6 is not supported by OpenZFS yet. In addition, if you are using kmods, you can ask DNF to list the contents of the ZFS package
```bash
dnf repoquery -l kmod-zfs
```
which clearly shows that the current ZFS kernel module does not support patch level 570.

Some of the issues linked in your bug report are quite irrelevant, since if you looked closely, you should see the latest OpenZFS version on RHEL 9.x remains on the 2.1.x branch. Whining about the 2.2.x branch not working on EL 9.6 when it is not even officially released into the EL 9.6 repo (which by the way doesn't even exist yet) is silly.

RHEL is stable within each minor release, but from one minor release to the next, the kernel API can change because of how Red Hat backports features and patches to their kernel, which breaks out of tree kernel modules. This is simply the nature of this release model. In my opinion, folks using EL in conjunction with third party kernel modules in production either have a Red Hat subscription and will be staying on the previous minor release using Extended Update Support (EUS) until such third party software has caught up, or they have a team that vets updates before rolling out upgrades. With this in mind, I don't think the relevant parts of the OpenZFS documentation is in any way inadequate.

macthecadillac:
From the docs, it is quite clear that OpenZFS is often behind on supporting new EL minor releases. In fact, if you try the command given in the documentation
```bash
dnf list available --showduplicates --releasever=9.6 kmod-zfs
```
or go to the repository at http://download.zfsonlinux.org/epel/, it is obvious that version 9.6 is not supported by OpenZFS yet. In addition, if you are using kmods, you can ask DNF to list the contents of the ZFS package
```bash
dnf repoquery -l kmod-zfs
```
which clearly shows that the current ZFS kernel module does not support patch level 570.

Some of the issues linked in your bug report seem irrelevant, since if you looked closely, you should see the latest OpenZFS version on RHEL 9.x for the moment remains on the 2.1.x branch. Whining about the 2.2.x branch not working on EL 9.6 when it is not even officially released into the EL 9.6 repo (which by the way doesn't even exist yet) is silly.

> We really just want to make sure future RHEL users are aware openzfs in its current state not only does not build on RHEL, but does not match RHEL users' expectations of a trouble-free, resilient file system they can simply throw on a box and call it a day.

It seems to me you have a fundamental misunderstanding of how the EL lifecycle works. EL is stable within each minor release, but from one minor release to the next, the kernel API can change because of how Red Hat backports features and patches to their kernel, which breaks out of tree kernel modules. This is simply the nature of this release model. This is very clearly explained in the docs:

> When updating to a new EL minor release the existing kmod packages may not work due to upstream kABI changes in the kernel. The configuration of the current release package may have already made an updated package available, but the package manager may not know to install that package if the version number isn’t newer. When upgrading, users should verify that the kmod-zfs package is providing suitable kernel modules, reinstalling the kmod-zfs package if necessary.

In my opinion, folks using EL in conjunction with third party kernel modules in production either have a Red Hat subscription and will be staying on the previous minor release using Extended Update Support (EUS) until such third party software has caught up, or they have a team that vets updates before rolling out upgrades. With this in mind, I don't think the relevant parts of the OpenZFS documentation is in any way inadequate. If you want to be able to just upgrade whenever updates arrive and expect things to simply work without having to deal with the EL lifecycle, I think you might be better off using a distribution that supports ZFS out of the box, such as Debian or Ubuntu.

svennd:
> > When updating to a new EL minor release the existing kmod packages may not work due to upstream kABI changes in the kernel. The configuration of the current release package may have already made an updated package available, but the package manager may not know to install that package if the version number isn’t newer. When upgrading, users should verify that the kmod-zfs package is providing suitable kernel modules, reinstalling the kmod-zfs package if necessary.

As a user, this could be more prominent in the docs. As I simply read over this.  We just upgrade and forget to validate this.

macthecadillac:
I suppose changing that block in the documentation from a note to a warning does not hurt. However, I still think it is the user's responsibility to understand the distribution they are using. In particular, that means understanding what Red Hat using semver on a distribution level truly means and how that release model is radically different from that of most other distributions.

While this sort of defeats the purpose of using an EL derivative, one workaround is to use a third party LTS kernel and use DKMS to install ZFS. The [CentOS kmods SIG](https://sigs.centos.org/kmods/repositories/) maintains a few LTS kernels in their repositories and they are, as far as I know, always up-to-date. There is also quite a popular set of [copr repositories maintained by kwizart](https://copr.fedorainfracloud.org/coprs/kwizart/) that provides up-to-date LTS kernels and they support EL, although from the download numbers it seems the main audience is Fedora.

PR #552, opened 2025-06-05
getting started: tell users to use partprobe

cyphar:
In almost all cases it is not necessary to reboot the machine in order
to reload the partition table for disks, you can just use partprobe(8)
which is part of parted and so is probably installed on all installation
media.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>


gmelikov (approved):
Thank you!

PR #558, opened 2025-06-20
Add hint for Debian Bookworm ZFS Root Networking

bdwalton:
Note that you may need to have appropriate firmware and networking utilities installed to have a working network upon reboot. This is varied so don't be overly specific but do point it out.

### DIFF
--- a/docs/Getting Started/Debian/Debian Bookworm Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bookworm Root on ZFS.rst
@@ -565,6 +565,16 @@ Step 4: System Configuration
 
    Customize this file if the system is not a DHCP client.
 
+#. Optional: Install driver firmware and WiFi support
+
+   If you're installing on a laptop or a device where wireless is the
+   primary network option, the above may not be sufficient as you
+   could lack the appopriate firmware for the device and tools to
+   configure the radio. Install some additional packages to cover
+   that need::
+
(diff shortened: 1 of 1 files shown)


rlaager:
This seems like a solid and helpful idea.

`firmware-linux` seems to be an actual (meta)package that depends on `firmware-linux-free` and `firmware-linux-nonfree`. So it seems like this could read, "You may need to install firmware-linux and other wireless network utilities as well."

Even better would be to specify the actual package name for the wireless utilities. Is this `wireless-tools`? (I don't personally use WiFi from CLI-only environments.)

Then, since we will have firm package names, maybe it should just be an optional step:

```
#. Optional: Install driver firmware and WiFi support

apt install --yes firmware-linux wireless-tools
```

Ideally, do you know what stock debian-installer installs? I like to match the results of the stock install as much as possible. If debian-installer installs e.g. firmware-linux, then we should just do that unconditionally when we install the kernel.

bdwalton:
I don't know what the stock installer does as I haven't used that path in years at this point as I always do zfs root and use this doc. This is the first time I've done it on a laptop though so I only noticed the need for this now. I think those two packages are likely sufficient. In my case, I just pulled in NetworkManager which pulled in everything else I needed and was on the path to what I wanted in the end anyway. I've updated the patch to reflect your suggestion.

gmelikov (approved):
Thank you!

Let's merge it as is for now, looks like a useful addition.

PR #559, opened 2025-06-25
Update Root on ZFS.rst

psy0rz:
mention alpinebox


docs/Getting Started/Alpine Linux/Root on ZFS.rst
```diff
@@ -30,6 +30,12 @@ follow instructions on this page if you plan to use ZBM,
 as the layouts are not compatible.  Refer
 to their site for installation details.
 
+**Alpinebox**
```
gmelikov, line 33: Proposal: let's wrap this and ZBM in "Alternatives to this guide" section and describe "you may use this or that as alternative to this guide" once.

PR #575, opened 2025-11-04
Update Feature Flags: illumos ZFS cannot read OpenZFS datasets with native encryption

wq9578:
> Due to incompatible formats, illumos ZFS cannot read OpenZFS datasets with native encryption.

This incompatibility should be mentioned to avoid unexpected results.

### DIFF
--- a/docs/Basic Concepts/Feature Flags.rst
+++ b/docs/Basic Concepts/Feature Flags.rst
@@ -23,6 +23,9 @@ Where all *features* that are used by a pool are supported by multiple
 implementations of OpenZFS, the on-disk format is portable across those
 implementations.
 
+Due to incompatible formats, illumos ZFS cannot read OpenZFS datasets
+with native encryption.
+
 Features that are exclusive when enabled should be periodically ported
 to all distributions.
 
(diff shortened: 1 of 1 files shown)


gmelikov:
> Maybe it is better to place the note in the general section instead of a footnote to only one filesystem, so users won't be surprised when starting with one filesystem and later find the footnote for the second system, not being able to use it then for their data.

So, we'll need to note about all features in this case? That's the whole purpose of features - different OSes have different features. I mean - you usually check "if this OS supports encryption feature?", and if it has "Yes^2" (note link) then you'll propably want to see notes. Did I miss something?

> The restriction of the statement should be placed right after: "Where all features that are used by a pool are supported by multiple implementations of OpenZFS, the on-disk format is portable across those implementations."

Indeed we may add note about "with explicit notes about support level if there's a difference".

What I mean - if we merge this PR as is - someday someone will inevitably miss such note and look only at feature matrix, and won't see this note.

> Also, I'm not familiar with Python, unfortunately.

Got you, I'll try to add it later, in a week or so. I'll add you as a coauthor, if you don't mind.

PR #580, opened 2025-11-20
Amend Debian package install command; update for Trixie.

ChibaPet:
Building on Trixie I noted that dh-dkms, libcurl4-openssl-dev, and
libtirpc-dev are required.

Also, package order passed to apt is irrelevant, so I sorted the list for 
easier consumption by humans.

### DIFF
--- a/docs/Developer Resources/Custom Packages.rst
+++ b/docs/Developer Resources/Custom Packages.rst
@@ -212,7 +212,7 @@ Make sure that the required packages are installed:
 
 .. code:: sh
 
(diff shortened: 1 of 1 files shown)


gmelikov:
Thank you! While most of changes are valid, I don't see that we need dh-dkms in our CI to build https://github.com/openzfs/zfs/blob/1f3444f2bba42817520a8097a29d24a3b4115927/.github/workflows/scripts/qemu-3-deps-vm.sh#L44 , will try to check later but I'd be glad if there's any info about that.

PR #584, opened 2025-11-28
added conf file to build readthedocs page

alek-p:
i've already configured http://openzfs.readthedocs.io/ to build from this repo.
I think adding this file is the last step needed to publish this docs repo at http://openzfs.readthedocs.io/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,29 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools we might need
+build:
+  os: ubuntu-24.04
+  tools:
+    python: "3.12"
+
+# Build documentation in the "docs/" directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+  # Fail on all warnings to avoid broken references
+  # fail_on_warning: true
+
+# Optionally build docs in additional formats such as PDF and ePub
+# formats:
+#   - pdf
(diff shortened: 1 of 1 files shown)


gmelikov:
My first thought is - add redirect from readthedocs to https://openzfs.github.io and call it a day, in this case we won't have additional service in dependencies (rtd), save all already working links and minimize the work. But it'd be great to look at other thoughts you originally had before that.

alek-p:
I don't see an issue with having both the GitHub.io page and the readthedocs one live at the same time and having the same info. This is what this PR would do and is the goal for me.

This is something that came up at the dev summit. The immediate problem we were trying to solve is to remove the old out of date docs from https://openzfs.readthedocs.io site and replace them with the ones we maintain as well as gaining control of the readthedocs page. There is too much old and or outdated info about zfs out there on the web and the readthedocs site has some authority in the open source community.

If we think we should only have one place for docs we could address that later on but I don't see the need to do that now.

gmelikov:
> I don't see an issue with having both the GitHub.io page and the readthedocs one live at the same time and having the same info. This is what this PR would do and is the goal for me.

It's harmful for indexing of our documentation in search engines (my past SEO experience screams, sorry), I've prepared redirection code here https://github.com/openzfs/openzfs-docs/pull/585

If you don't mind redirects from RTD to https://openzfs.github.io/ - it should work. Otherwise - we can change main domain in future, but it's better to have one main domain at a time.

PS: mirror pages on rtd may have problems with `/openzfs-docs/` prefix, jfyi, it may even give some problems with my redirection code, but I can easily fix it if there's a mechanism to deliver code there.

PR #587, opened 2025-11-30
Use sphinx>=8 and disable shallow clone in CI

gmelikov:
It's needed to fix sphinx_last_updated_by_git in CI.


gmelikov:
In addition - ammaraskar/sphinx-action CI dependency was dropped, we already have ready to use sphinx in system, and we need `git` for sphinx build now. I've checked for good resulting html in CI artifacts https://github.com/openzfs/openzfs-docs/actions/runs/19804487003/artifacts/4718398059

alek-p (approved):

PR #592, opened 2025-12-16
Add backporting instructions

tonyhutter:
We basically have no written policy on how we want backports to be done.  This came up as an issue in https://github.com/openzfs/zfs/pull/18038.  We should document what we are currently doing so it's not ambiguous.


mabod:
This PR is talking about "backporting" but the commit starts with the line "**Backpointing** a commit to a release branch". Looks like a typo.

tonyhutter:
@mabod good catch - I fixed the typo in my latest push.

PR #596, opened 2026-01-08
LLM: add man pages + Module parameters: add note about #491 issue

gmelikov:
Example: https://github.com/openzfs/openzfs-docs/actions/runs/20833409146/artifacts/5069172946


gmelikov:
@alek-p offtopic - now llms-full.txt will be here https://openzfs.github.io/openzfs-docs/llms-full.txt . Is it ok or where usually it should be? How will LLMs find it? Google says there's some sites-`indexes` which can have non-root paths.

docs/Performance and Tuning/Module Parameters.rst
```diff
@@ -1,6 +1,14 @@
+.. this file is not actual, actual info will be added to context from man pages.
+
+:llms-txt-ignore: true
+
 Module Parameters
 =================
 
+.. note::
+   This page is pretty obsolete and doesn't reflect all parameters,
```
alek-p: I actually like and use this page. I think it's valuable to have a curated list of useful tunables. What do you think about rephrasing this to somethink like:

This page is not actively maintained and is incomplete, see issue https://github.com/openzfs/openzfs-docs/issues/491.
For a complete list of paramiters see https://openzfs.github.io/openzfs-docs/man/master/4/index.html

docs/Performance and Tuning/Module Parameters.rst
```diff
@@ -1,6 +1,14 @@
+.. this file is not actual, actual info will be added to context from man pages.
+
+:llms-txt-ignore: true
```
alek-p: what does this do?

alek-p (changes requested):

docs/Performance and Tuning/Module Parameters.rst
gmelikov: Problem is - it's not incomplete, it's not actual in the first place, there are many parameters which don't exist at all in master.

And it's mainly a duplicate of man pages, it adds 10k lines to llms-full.txt (which is ~60k lines already). 

AFAIK largest problem for LLM is a context size now, so that's why I try to remove non-unique data from it.

gmelikov:
https://llmstxt.org/ gives examples on https://www.fastht.ml/docs/llms.txt and it's not in their sitemap.xml, so let's try as is.

docs/Performance and Tuning/Module Parameters.rst
gmelikov: Updated with different wording, please look.

PR #601, opened 2026-01-26
docs: Workload Tuning: Expand PostgreSQL section

### LINKED ISSUE #598: Motivate PostgreSQL recordsize recommendations
The Tuning Guide [currently says](https://github.com/openzfs/openzfs-docs/blob/5ad3a5ff6227a499f576d886592692fd59b7d6c5/docs/Performance%20and%20Tuning/Workload%20Tuning.rst#postgresql) for postgres:

> Set `compression=lz4` and `recordsize=32K` (64K also work well, as does the 128K default)

I would first think that `recordsize=8K` should be optimal given that postgres pages are 8K. Especially because the section on SQLite just below it says

> Set the recordsize to 64K. Set the SQLite page size to 65536 bytes

which matches DB and ZFS record sizes.

Probably the recommendation for larger records is because of `compression`? It would be great if that section could explicitly point it out.

---

Relatedly, I found that the `man zfsprops` info on DBs looked very wrong, perhaps good to cross-check that:

* https://github.com/openzfs/zfs/issues/18156

Thanks!

nh2:
Closes #598.

### DIFF
--- a/docs/Performance and Tuning/Workload Tuning.rst
+++ b/docs/Performance and Tuning/Workload Tuning.rst
@@ -652,9 +652,23 @@ PostgreSQL
 
 Make separate datasets for PostgreSQL's data and WAL. Set 
 ``compression=lz4`` and ``recordsize=32K`` (64K also work well, as 
-does the 128K default) on both. Configure ``full_page_writes = off`` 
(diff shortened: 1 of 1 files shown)


nh2:
Thanks!

PR #631, opened 2026-04-18
Fix dead tarball URL and http→https in message example

chrislongros:
## Summary

- `Custom Packages.rst` pointed to `archive.zfsonlinux.org`, which no longer responds. Updated to the current GitHub Releases URL.
- `msg/ZFS-8000-ER/index.rst` example showed an `http://` self-reference; `zpool` itself prints `https://` today. Updated to match.

## Testing 

- [x] No build regressions (Sphinx-only changes)
- [x] Verified the new tarball URL resolves
- [x] Verified the `https://` form matches what `cmd/zpool/zpool_main.c` prints

### DIFF
--- a/docs/Developer Resources/Custom Packages.rst
+++ b/docs/Developer Resources/Custom Packages.rst
@@ -304,7 +304,7 @@ then use the following commands to fetch and prepare the source.
 
 .. code:: sh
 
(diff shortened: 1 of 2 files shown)


gmelikov (approved):
Thank you!

PR #632, opened 2026-05-11
Add Secure Boot warning to Debian, Ubuntu, and NixOS root guides

### LINKED ISSUE #493: SecureBoot and zfs kernel module
While following [Debian Bookworm Root on ZFS](https://openzfs.github.io/openzfs-docs/Getting%20Started/Debian/Debian%20Bookworm%20Root%20on%20ZFS.html) the setup of `zfs-dkms` fails to load the kernel module after the build step, printing:

    insmod /lib/modules/6.1.0-15-amd64/updates/dkms/spl.ko                                                                           
    modprobe: ERROR: could not insert 'zfs': Key was rejected by service

A user installing Debian Bookworm Root on ZFS may not notice this, and only come to a halt at the first `zpool create` which returns with:

    The ZFS modules are not loaded.                                                                                                  
    Try running '/sbin/modprobe zfs' as root to load them.

Of course, attempting `/sbin/modprobe zfs` comes back with the same:

    modprobe: ERROR: could not insert 'zfs': Key was rejected by service
    
The solution to this was to change the SecureBoot setting in the BIOS from `Deployed Mode` to `Audit Mode` . This should be presented as a warning and instruction at the start of the procedure, to avoid having to reboot and setup the install environment again.

chrislongros:
Other Root-on-ZFS guides already note that ZFS modules cannot load with Secure Boot enabled (Fedora, Alpine, RHEL-based, Arch). The Debian,Ubuntu, and NixOS guides did not.

Closes #493

Signed-off-by: Christos Longros <chris.longros@gmail.com>

### DIFF
--- a/docs/Getting Started/Debian/Debian Bookworm Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bookworm Root on ZFS.rst
@@ -104,6 +104,7 @@ encrypted once per disk.
 Step 1: Prepare The Install Environment
 ---------------------------------------
 
+#. Disable Secure Boot. ZFS modules can not be loaded if Secure Boot is enabled.
 #. Boot the Debian GNU/Linux Live CD. If prompted, login with the username
    ``user`` and password ``live``. Connect your system to the Internet as
    appropriate (e.g. join your WiFi network). Open a terminal.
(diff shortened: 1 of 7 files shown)


gmelikov:
IIRC it can work with secure boot if you make some additional steps with MOK and self-sign modules https://wiki.debian.org/SecureBoot#MOK_-_Machine_Owner_Key (see `DKMS and Secure Boot` there too).

Didn't test it lately though, maybe we should add something about it in warning text.

rlaager:
With Ubuntu, shouldn't this be fine, since the modules ship with the upstream kernel?

Granted, I'm way behind in updating this, but hopefully I'll be getting back to this for Ubuntu 26.04 shortly. Ideally, I should update the instructions for Debian to use a local signing key.

PR #637, opened 2026-07-21
Debian: move optional WiFi/firmware step into the chroot + misc

### LINKED ISSUE #635: Incorrect order for instructions about installing wifi prerequisites?
I am using the instructions for ZFS Root on Debian trixie:

https://openzfs.github.io/openzfs-docs/Getting%20Started/Debian/Debian%20Trixie%20Root%20on%20ZFS.html#step-4-system-configuration

Under Step 4, System Configuration, point 3 is:

> 3. Optional: Install driver firmware and WiFi support
>
> If you’re installing on a laptop or a device where wireless is the primary network option, the above may not be sufficient as you could lack the appropriate firmware for the device and tools to configure the radio. Install some additional packages to cover that need

This is indeed something I needed! However, I believe this step comes too soon, as running the commands at this point installs packages in the Debian live system (and in fact they're already installed there!), not in the target system.

I believe that this point 3 should come after point 6 ("Configure a basic system environment"), when we are now in the chroot and running on the target system, but still with the live CD's kernel and hence WiFi drivers (so it's not too late, as it would be if we waited until First Boot).

(Many thanks again for these HOWTOs. I don't believe I've ever used a better HOWTO, even in the golden age of such documents from 20 or 30 years ago! I find the level of detail to be exactly what is required: I find the answers to every *essential* question in the text, so I'm reassured that the instructions are correct and I feel that I understand what I'm doing; while of course these days I can easily look up material that is not strictly necessary or relevant to the task. I also appreciate the level of care and detail; for example, the instructions explain that in Trixie specifically it is necessary to install an ntp server. The fact that the HOWTOs are kept up to date is also very helpful and reassuring.)

gmelikov:
The 'Install driver firmware and WiFi support' step ran 'apt install' while still in the live environment, so the packages landed in (or were already present in) the live system instead of the target system.

Move it after 'Configure a basic system environment', which is inside the chroot but still before the live CD environment is left, so the live CD's WiFi drivers remain available.

Applies to both Bookworm and Trixie guides.

Closes: #635

### DIFF
--- a/docs/Getting Started/Debian/Debian Bookworm Root on ZFS.rst
+++ b/docs/Getting Started/Debian/Debian Bookworm Root on ZFS.rst
@@ -573,16 +573,6 @@ Step 4: System Configuration
 
    Customize this file if the system is not a DHCP client.
 
-#. Optional: Install driver firmware and WiFi support
-
-   If you're installing on a laptop or a device where wireless is the
-   primary network option, the above may not be sufficient as you
-   could lack the appropriate firmware for the device and tools to
-   configure the radio. Install some additional packages to cover
-   that need::
-
-     apt install --yes firmware-linux wireless-tools
-
 #. Configure the package sources::
 
      vi /mnt/etc/apt/sources.list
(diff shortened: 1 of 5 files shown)


rrthomas:
Thanks very much for this; sorry I didn't get to it before you did!

PR #644, opened 2026-08-01
Update Debian packages

PeterBeu:
I think these were Ubuntu package names. This is all that's actually needed on Debian. Tested on a clean install of trixie.


gmelikov (approved):
Thank you!

